PR #4, opened 2022-05-13
pseduo code

WillBeebe:


pkg/compress/tar.go
```diff
@@ -42,13 +42,46 @@ func TarDirectory(dirPath string, prefix string, tarWriter *tar.Writer) error {
 			if _, err := io.Copy(tarWriter, data); err != nil {
 				return err
 			}
+		} else if fi.IsDir() && fi.Name() == "chart" {
+			packageChart(path.Join(path.Dir(appPath), app.Deployment.Path), path.Join(workingDir, "chart"))
```
WillBeebe, line 46: I know this method won't work here, but we basically want this functionality.

The code below is a rough attempt at reading and writing the files to the tarWriter but I don't know how to do directories.

PR #5, opened 2022-05-14
Will/support local charts in bundle

WillBeebe:
I tested this with the Shoptracker mongo bundle and we're green! I'm trying to write a test for it and struggling a bit, so I didn't include that here. Current tests pass.

### DIFF
--- a/pkg/compress/tar.go
+++ b/pkg/compress/tar.go
@@ -86,7 +86,7 @@ func TarFile(filePath string, destinationPrefix string, tarWriter *tar.Writer) e
 	}
 
 	if fileInfo.IsDir() {
-		return errors.New("specified path is not a absolueSourceFilePath")
+		return errors.New("specified path is not a file")
 	}
 
 	header, err := tar.FileInfoHeader(fileInfo, filePath)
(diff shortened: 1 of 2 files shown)


pkg/compress/tar.go
```diff
@@ -10,46 +10,70 @@ import (
...
-	filepath.Walk(absolutePath, func(file string, fi os.FileInfo, err error) error {
-		// generate tar header
-		header, err := tar.FileInfoHeader(fi, file)
+	// walk through every absolueSourceFilePath in the folder
+	filepath.Walk(absoluteSourcePath, func(absolueSourceFilePath string, info os.FileInfo, err error) error {
+		relativeDestinationFilePath := path.Join(destinationPrefix, strings.TrimPrefix(filepath.ToSlash(absolueSourceFilePath), absoluteSourcePath))
+
+		// this feels a little weird, "ideally" we pass in the relativeSourceFilePath
+		// however, we'd only use that here and it would only make this test slightly more sane
+		if info.IsDir() || ShouldIgnore(relativeDestinationFilePath) {
```
WillBeebe, line 25: this seems to very important. When I was trying to write the _folder_, the file writer of course doesn't work. I did some digging and I don't think we want to add a header to the tar for the folder, if we use the relative path for every _file_ un-tarring will get us the folder structure.

The `ShouldIgnore` method was also important!  It was uploading my `.terraform` , `connections.tfvars.jsson`, etc.. I hacked together a prefix match to ignore these files, lmk what you think.

PR #18, opened 2022-05-23
App envs

WillBeebe:
Terraform expects `envs`, `envs` is better for an array than `env`. This is the last thing to make app dependencies work again. 🥳


coryodaniel (approved):
Makes sense to me as an array

PR #28, opened 2022-06-05
during bundle generate, ask the user where they want to output the bu…

chrisghill:
…ndle


cmd/application.go
```diff
@@ -17,15 +17,17 @@ import (
 
 // applicationCmd represents the application command
 var applicationCmd = &cobra.Command{
-	Use:   "application",
-	Short: "Application development tools",
-	Long:  ``,
+	Use:     "application",
+	Aliases: []string{"app"},
```
chrisghill, line 21: just adding an alias here, so you can do `mass app gen`

pkg/bundle/prompt.go
```diff
@@ -86,3 +87,19 @@ func getDescription(t *TemplateData) error {
 	t.Description = result
 	return nil
 }
+
+func getOutputDir(t *TemplateData) error {
```
chrisghill, line 91: This is the only real change associated with the prompt

pkg/bundle/generate.go
```diff
@@ -1,127 +1,50 @@
 package bundle
```
chrisghill, line 1: These changes are a red herring.  I got rid of the `generator` package and moved the files into the `bundle` package.  However, both packages had a `generate.go` file. So I moved the `bundle/generate.go` to `bundle/schema.go`, and the `generator/generate.go` became `bundle.generate.go`.

So basically GitHub is throwing a ton of diffs when its a completely different file.

pkg/bundle/generate_test.go
```diff
@@ -1,49 +1,56 @@
 package bundle_test
```
chrisghill, line 1: Same as above. This file is now `bundles/schema_test.go` and `generator/generate_test.go` is this file. The diff here is irrelevant.

pkg/bundle/schema.go
```diff
@@ -0,0 +1,127 @@
+package bundle
```
chrisghill, line 1: This was `bundle/generator.go`

pkg/bundle/schema_test.go
```diff
@@ -0,0 +1,49 @@
+package bundle_test
```
chrisghill, line 1: This was `generator_test.go`. No changes were made.

pkg/bundle/prompt.go
```diff
@@ -86,3 +87,19 @@ func getDescription(t *TemplateData) error {
...
+	prompt := promptui.Prompt{
+		Label:   `Output directory`,
+		Default: t.Name,
```
WillBeebe, line 94: perfect default 👌

WillBeebe (approved):
LGTM thanks Chris!

PR #30, opened 2022-06-08
Default steps if not present

coryodaniel:
We should default the steps to `src` and `terraform` when not present in the file and exclude them from the bundle generator.

This will let us hide this feature from users for now so people don't depend on it since it can be destructive in the even they change their `path`.

It also makes the standard bundle file a little shorter for something the average users won't need to change.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,21 @@
+name: Test
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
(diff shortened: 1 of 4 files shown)


WillBeebe (approved):
Seems reasonable. I haven't run into this yet because after using the cli to scaffold, we've been copy-pasting the `massdriver.yaml`. Thanks for finding / fixing.

jaketf (approved):

chrisghill (approved):

PR #32, opened 2022-06-08
Cmd line override of access

coryodaniel:
Enable the ability to override `access` on `mass bundle publish`.

This will allow us in CI to override the value so we can keep sandbox bundles private and official bundles public.

### DIFF
--- a/pkg/application/package.go
+++ b/pkg/application/package.go
@@ -16,8 +16,6 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-// TODO: we need to add support for access overrides to applications as well
-// Remove comment in PR via GH suggestion and open issue.
 func PackageApplication(appPath string, c *client.MassdriverClient, workingDir string, buf io.Writer) (*bundle.Bundle, error) {
 	app, err := Parse(appPath)
 	if err != nil {
--- a/pkg/application/package.go
+++ b/pkg/application/package.go
@@ -16,6 +16,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+// TODO: we need to add support for access overrides to applications as well
+// Remove comment in PR via GH suggestion and open issue.
 func PackageApplication(appPath string, c *client.MassdriverClient, workingDir string, buf io.Writer) (*bundle.Bundle, error) {
 	app, err := Parse(appPath)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/bundle/parse.go
```diff
@@ -20,5 +26,17 @@ func Parse(path string) (*Bundle, error) {
 		return nil, err
 	}
 
+	applyOverrides(bundle, overrides)
```
WillBeebe, line 29: Do we need a guard for `nil` overrides?

https://github.com/massdriver-cloud/massdriver-cli/pull/32/files#diff-409246a696ba52eb48e1b714110694f2e0ff050f72b06a4e0eefdfbd76e0ca5dR89

pkg/bundle/parse_test.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func TestParseBundle(t *testing.T) {
-	var got, _ = bundle.Parse("./testdata/massdriver.yaml")
+	var got, _ = bundle.Parse("./testdata/massdriver.yaml", nil)
```
WillBeebe, line 11: answered my question with a test 💯

pkg/bundle/parse.go
coryodaniel: its in `applyOverrides` with a comment that we need to be actually validating the bundles w/ a metaschema. its a pedantic `if ||` check today.

pkg/application/package.go
```diff
@@ -16,6 +16,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+// TODO: we need to add support for access overrides to applications as well
+// Remove comment in PR via GH suggestion and open issue.
```
coryodaniel, line 20:
```suggestion
```

https://github.com/massdriver-cloud/massdriver-cli/issues/33

WillBeebe (approved):
LGTM

pkg/bundle/parse.go
```diff
@@ -20,5 +26,17 @@ func Parse(path string) (*Bundle, error) {
...
+
 	return bundle, nil
 }
+
+func applyOverrides(b *Bundle, overrides map[string]interface{}) {
+	if access, found := overrides["access"]; found {
+		// TODO: we need to add a meta schema for our metadata and validate the bundle
```
coryodaniel, line 36:
```suggestion
```
https://github.com/massdriver-cloud/massdriver-cli/issues/34

PR #39, opened 2022-06-24
Fix the canonical issue with JSON schema validation and add verbose l…

chrisghill:
closes: https://github.com/massdriver-cloud/massdriver/issues/1055

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -7,6 +7,7 @@ package cmd
 import (
 	"os"
 
+	"github.com/rs/zerolog"
 	"github.com/spf13/cobra"
 )
 
@@ -43,4 +44,15 @@ func init() {
 	// when this action is called directly.
 	rootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
 	rootCmd.PersistentFlags().StringP("api-key", "k", "", "Massdriver API key (can also be set via MASSDRIVER_API_KEY environment variable)")
+	rootCmd.PersistentFlags().BoolP("verbose", "v", false, "Enable verbose output")
+}
+
+func setupLogging(cmd *cobra.Command) {
+	verbose, _ := cmd.Flags().GetBool("verbose")
+
+	if verbose {
+		zerolog.SetGlobalLevel(zerolog.DebugLevel)
+	} else {
+		zerolog.SetGlobalLevel(zerolog.WarnLevel)
+	}
 }
--- a/cmd/schema.go
+++ b/cmd/schema.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/massdriver-cloud/massdriver-cli/pkg/jsonschema"
@@ -41,13 +42,28 @@ func init() {
 }
 
 func runSchemaValidate(cmd *cobra.Command, args []string) error {
+	setupLogging(cmd)
+
 	schema, _ := cmd.Flags().GetString("schema")
 	document, _ := cmd.Flags().GetString("document")
-	_, err := jsonschema.Validate(schema, document)
-	return err
+	valid, violations, err := jsonschema.Validate(schema, document)
+	if err != nil {
+		return err
+	}
+	if valid {
+		fmt.Println("The document is valid!")
+	} else {
+		fmt.Printf("The document failed validation:\n\tDocument: %s\n\tSchema: %s\nErrors:\n", document, schema)
+		for _, violation := range violations {
+			fmt.Printf("\t- %v\n", violation)
+		}
+	}
+	return nil
 }
 
 func runSchemaDereference(cmd *cobra.Command, args []string) error {
+	setupLogging(cmd)
+
 	schema := args[0]
 	out, _ := cmd.Flags().GetString("out")
 
--- a/cmd/schema.go
+++ b/cmd/schema.go
@@ -37,6 +37,7 @@ func init() {
 
 	schemaValidateCmd.Flags().StringP("document", "d", "document.json", "Path to JSON document")
 	schemaValidateCmd.Flags().StringP("schema", "s", "./schema.json", "Path to JSON Schema")
+	schemaValidateCmd.Flags().BoolP("exit-error", "e", false, "Exit non-zero if the schema is invalid")
 
 	schemaDereferenceCmd.Flags().StringP("out", "o", "", "File to output derefenced schema to (default is stdout)")
 }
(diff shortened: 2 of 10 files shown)


cmd/root.go
```diff
@@ -43,4 +44,15 @@ func init() {
 	// when this action is called directly.
 	rootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
 	rootCmd.PersistentFlags().StringP("api-key", "k", "", "Massdriver API key (can also be set via MASSDRIVER_API_KEY environment variable)")
+	rootCmd.PersistentFlags().BoolP("verbose", "v", false, "Enable verbose output")
+}
+
+func setupLogging(cmd *cobra.Command) {
```
chrisghill, line 50: Adding this so we have a `-v` flag to toggle showing our JSON logs.  Should make the CLI appear more polished to users when we don't throw a bunch of JSON at them every time they run a command.

pkg/jsonschema/loader.go
```diff
@@ -16,6 +17,10 @@ func Loader(path string) gojsonschema.JSONLoader {
 	if loaderPrefixPattern.MatchString(path) {
 		ref = path
 	} else {
+		// gojsonschema has a strange "reference must be canonical" error if the schema path is the current directory
```
chrisghill, line 20: Super weird bug here. If the schema file exists in the current directory, and you don't prefix the path with `./` (`./schema.json` vs `schema.json`) it will fail with a cryptic "schema must be canonical" error.

https://github.com/xeipuuv/gojsonschema/issues/92

This seems to fix the issue.

pkg/jsonschema/validate.go
```diff
@@ -1,38 +1,33 @@
...
-	"errors"
-	"fmt"
-
 	"github.com/rs/zerolog/log"
 	"github.com/xeipuuv/gojsonschema"
 )
 
 // Validate the input object against the schema
-func Validate(schemaPath string, documentPath string) (bool, error) {
+func Validate(schemaPath string, documentPath string) (bool, []string, error) {
```
chrisghill, line 9: Turned this into more of a library function. It now returns a string list of the violations it finds, allowing the caller to do with them whatever is appropriate (as opposed to always printing to screen).

Technically the boolean return value is a bit moot now, since we can imply "valid" or "invalid" based on the length of the violations list (`len == 0` implies valid). I left it in since it seems convenient, but more importantly I don't trust that there will be an edge case where `gojsonschema` returns `valid = false` but there are 0 results.

cmd/schema.go
```diff
@@ -41,13 +42,28 @@ func init() {
...
+
 	schema, _ := cmd.Flags().GetString("schema")
 	document, _ := cmd.Flags().GetString("document")
-	_, err := jsonschema.Validate(schema, document)
-	return err
+	valid, violations, err := jsonschema.Validate(schema, document)
+	if err != nil {
+		return err
+	}
+	if valid {
```
chrisghill, line 53: Moved the printing out of the library and up to here.

Also added a explicit message when the document is valid.

chrisghill:
Before:
```
$ mass schema validate -s schema-params.json -d src/dev.params.tfvars.json
{"level":"debug","schemaPath":"schema-params.json","documentPath":"src/dev.params.tfvars.json","time":"2022-06-24T00:43:08-06:00","message":"Validating schema."}
{"level":"error","error":"Reference file://schema-params.json must be canonical","time":"2022-06-24T00:43:08-06:00","message":"Validator failed."}
Error: Reference file://schema-params.json must be canonical
Usage:
  mass schema validate [flags]

Flags:
  -d, --document string   Path to JSON document (default "document.json")
  -h, --help              help for validate
  -s, --schema string     Path to JSON Schema (default "./schema.json")

Global Flags:
  -k, --api-key string   Massdriver API key (can also be set via MASSDRIVER_API_KEY environment variable)
```

After:
```
$ mass schema validate -s schema-params.json -d src/dev.params.tfvars.json
Document is valid!
```

cmd/root.go
coryodaniel: We should update the output to not be json at some point since it’s a cli and to be consumed by a human

pkg/jsonschema/validate.go
```diff
@@ -1,38 +1,33 @@
...
 		}
 
-		err = errors.New(msg)
-		log.Error().Err(err).Msg("Validation failed.")
-		return false, err
+		return false, violations, err
 	}
 
-	return true, nil
+	return true, violations, nil
```
coryodaniel, line 32: A three arity return function feels a bit weird. You can kind of see this in the jsonschema library w the result struct containing Valid and Errors. 

With the actual error, how is that ever handled by a caller? Does it bundle up and just eventually panic? If so it might make sense to exit w an error to the user here

chrisghill: Agreed, I don't like the 3 values.

I'd rather just return the `gojsonschema` result object and let the caller unpack it. I don't like the idea of conflating a runtime execution error with what is effectively a "false" condition check. For example, I think there is a difference between an error caused by an invalid file path vs everything ran just fine, but the document isn't valid. And I don't think it should be the responsibility of the caller to figure out which is which.

coryodaniel: I think that’s valid if we exit non-zero with whatever the unrecoverable error is

pkg/jsonschema/validate.go
```diff
@@ -1,38 +1,18 @@
...
-			msg = msg + fmt.Sprintf("\t- %s\n", desc)
-		}
-
-		err = errors.New(msg)
-		log.Error().Err(err).Msg("Validation failed.")
-		return false, err
-	}
-
-	return true, nil
+	return gojsonschema.Validate(sl, dl)
```
chrisghill, line 17: @coryodaniel Just returning the result from validate now 👍

cmd/schema.go
```diff
@@ -36,18 +37,39 @@ func init() {
 
 	schemaValidateCmd.Flags().StringP("document", "d", "document.json", "Path to JSON document")
 	schemaValidateCmd.Flags().StringP("schema", "s", "./schema.json", "Path to JSON Schema")
+	schemaValidateCmd.Flags().BoolP("exit-error", "e", false, "Exit non-zero if the schema is invalid")
```
chrisghill, line 40: @coryodaniel Added this useful flag for when this is run in automation. By default it will return 0 even if the schema is invalid, but enabling this will cause it to return 1 if its invalid.

PR #40, opened 2022-07-06
add validations.json to bundle generation

jaketf:
I added this new generated file in the `src/` directory. In theory there could be collision between address space in a bundle steps because they are separate tf states. @chrisghill lmk if this file is expected at top level (next to massdriver.yaml) should be simple to move it.

during this PR I noticed a lot of the files in this repo don't obey the pre-commit and we could use to turn on some more aggressive go linting. I'lll propose those changes in a separate PR.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,5 +25,6 @@ repos:
       - id: check-yaml
       - id: end-of-file-fixer
       - id: detect-aws-credentials
+        args: [--allow-missing-credentials]
       - id: no-commit-to-branch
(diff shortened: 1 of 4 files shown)


chrisghill (approved):
Looks great. I like the cleanup as well!

chrisghill:
> I added this new generated file in the src/ directory. In theory there could be collision between address space in a bundle steps because they are separate tf states.
The `validations.json` goes in the `src/` directory. Since each bundle step is in its own directory, it would also have its own `validations.json` file to prevent collisions 👍

PR #41, opened 2022-07-06
app dependencies to map instead of array

WillBeebe:
Grabbed this at lunch today. 

before
```yaml
dependencies:
  - type: massdriver/type
     field: database
```

after
```yaml
dependencies:
  database:
    type: massdriver/type
```

PR #44, opened 2022-07-07
enable and abide by initial set of golangci-lint rules

jaketf:
This PR does the following to codify our golang coding standards and keep us honest in this repo:
- runs pre-commit in CI 
- enables a slew of `golangci-lint` linters (you can learn more [here](https://golangci-lint.run/)) I loosely populated out config based on [this example](https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322) there's plenty more we could enable and we can develop more sophisticated opinions over time.
- I opted to silence some linter warning with a `// nolint: xxx` comment in the name of getting this already huge PR out the door. We can hopefully get around to removing as many of those as possible in the future. 

### High level explanation of changes
- Making functions more readable: you'll see a lot of jake moving small bits of code into their own functions, the reason is the higher level orchestrating function becomes more easily readable because the smaller function names narrate what's going on, this makes it easier to get to the piece of code you're trying to change / debug. 
  - functions should not be more than 50 statements (funlen)
  - functions should not be very complex (many nested if statement logics)
- there's a lot of changes that have to do with no shadowing the name `err` this makes the code a lot easier to re-order / delete without having to figure out where err was defined
- DRY up `const`s (e.g. massdriver ui)
- follow Golang naming convention: all acronyms should be all caps not CamelCase (e.g. API not Api)
- avoid [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) 
- start to introduce the use of contexts. We can get more elaborate with this to be able to set timeouts, rate limits, or cancel requests when / if the cli gets more complex. It is generally considered best practice to "every time you make an API call there should be a context associated with that. I'm honestly not sure I defined contexts at the "proper level in the call stack" in this PR but at least we can start with this. 
- catch. every. error. 
- pretty format generated json
  - not a go change but i also pretty formatted all the json in this repo so we actually obey the end of file fixer in the pre-commit
- make changes to dev container so the pre-commit / golangci-lint can run in the dev container (would be great if a reviewer could validate this on their machine)


.pre-commit-config.yaml
```diff
@@ -5,13 +5,13 @@ repos:
...
       - id: go-vet
-      #- id: go-lint
+      # - id: go-lint
       #- id: go-imports
       #- id: go-cyclo
       #  args: [-over=15]
       #- id: validate-toml
       - id: no-go-testing
-      #- id: golangci-lint
+      - id: golangci-lint
```
WillBeebe, line 14: 👌

pkg/definition/get.go
```diff
@@ -17,7 +18,9 @@ func GetDefinition(c *client.MassdriverClient, definitionType string) (map[strin
 
 	req := client.NewRequest("GET", endpoint, nil)
 
-	resp, err := c.Do(req)
+	ctx := context.TODO()
```
WillBeebe, line 21: nice use of the `context.TODO()` in the PR

cmd/definition.go
```diff
@@ -44,7 +44,9 @@ func init() {
 
 	definitionCmd.AddCommand(definitionPublishCmd)
 	definitionPublishCmd.Flags().StringP("file", "f", "", "File containing artifact definition schema (use - for stdin)")
-	definitionPublishCmd.MarkFlagRequired("file")
+	if err := definitionPublishCmd.MarkFlagRequired("file"); err != nil {
```
WillBeebe, line 47: I didn't know you could do this with if statements, cool!

cmd/definition.go
```diff
@@ -109,11 +111,13 @@ func runDefinitionPublish(cmd *cobra.Command, args []string) error {
...
-	json.Unmarshal([]byte(byteValue), &art)
+	if jsonErr := json.Unmarshal(byteValue, &art); err != nil {
+		return jsonErr
+	}
 
-	err = art.Publish(c)
-	if err != nil {
-		return err
+	pubErr := art.Publish(c)
+	if pubErr != nil {
```
WillBeebe, line 119: style-wise, when do we keep these separated vs

```golang
if pubErr:= art.Publish(c); pubErr != nil {
  return pubErr
}
```

jaketf: good question and this one would actually go better the way you suggest!

the reason to separate is for two value returns (because go is very strict about the scope of the values in an if statement remaining in that closure).

pkg/provisioners/terraform/compile_test.go
```diff
@@ -1,21 +1,25 @@
...
+
+	"github.com/massdriver-cloud/massdriver-cli/pkg/provisioners/terraform"
 )
 
 // Helper function for asserting json serde matches
 func doc(str string) string {
 	b := []byte(str)
 
 	jsonMap := make(map[string](interface{}))
-	json.Unmarshal([]byte(b), &jsonMap)
```
jaketf, line 16: FYI to reviewer this used to make the "empty shema" `doc("")` silently fail. adding this error handling is why I needed to make `doc("{}")` change. 

@chrisghill hopefully this isn't tangibly different in the sentiment of what's being tested. LMK if this needs to be handled differently.

cmd/definition.go
jaketf: longer better answer https://www.calhoun.io/one-liner-if-statements-with-errors/

pkg/application/convert_test.go
```diff
@@ -22,18 +22,18 @@ func TestConvertToBundle(t *testing.T) {
 				Description: "description",
 				Ref:         "github.com/some-repo",
 				Access:      "public",
-				Dependencies: []application.ApplicationDependencies{
+				Dependencies: []application.Dependencies{
```
chrisghill, line 25: FYI, @WillBeebe just made a change to this where the Dependencies becomes a `map[string]application.Dependencies`, and we drop `field` from the struct.

https://github.com/massdriver-cloud/massdriver-cli/pull/41

Not a complex change, but one of you will have a merge conflict.

pkg/application/generate.go
```diff
@@ -63,7 +62,7 @@ func Generate(data *ApplicationTemplateData) error {
 	if err != nil {
 		return err
 	}
-	err = ioutil.WriteFile(path.Join(tempDir, data.Chart, "Chart.yaml"), chartBytes, 0644)
+	err = ioutil.WriteFile(path.Join(tempDir, data.Chart, "Chart.yaml"), chartBytes, 0644) // nolint:gosec,gomnd
```
chrisghill, line 65: I'm assuming this is a comment telling the linter to skip `gosec` and `gomnd`. What do they not like about this line?

pkg/application/prompt.go
```diff
@@ -33,15 +33,15 @@ func RunPrompt(t *ApplicationTemplateData) error {
...
+func getName(t *TemplateData) error {
 	validate := func(input string) error {
 		if !bundleTypeFormat.MatchString(input) {
 			return errors.New("name must be 2 or more characters and can only include lowercase letters and dashes")
 		}
 		return nil
 	}
 
-	defaultValue := strings.Replace(strings.ToLower(t.Name), " ", "-", -1)
+	defaultValue := strings.ReplaceAll(strings.ToLower(t.Name), " ", "-")
```
chrisghill, line 44: Good catch!

pkg/bundle/hydrate.go
```diff
@@ -9,43 +10,44 @@ import (
...
 	cwd := filepath.Dir(path)
+	ctx := context.TODO()
 
-	hydratedArtifacts, err := jsonschema.Hydrate(b.Artifacts, cwd, c)
+	hydratedArtifacts, err := jsonschema.Hydrate(ctx, b.Artifacts, cwd, c)
 	if err != nil {
 		return err
 	}
-	b.Artifacts = hydratedArtifacts.(map[string]interface{})
+	b.Artifacts = hydratedArtifacts.(map[string]interface{}) //nolint:errcheck
```
chrisghill, line 19: Does the linter not like type assertions?

pkg/common/constants.go
```diff
@@ -0,0 +1,6 @@
+package common
+
+const (
+	MassdriverURL = "https://api.massdriver.cloud/"
+	HTTPCodeOK    = 200
```
chrisghill, line 5: We probably don't need to define our own constant for this. The `http` pkg has `http.StatusOK` if we want to use a constant.

pkg/jsonschema/hydrate.go
```diff
@@ -19,104 +20,136 @@ var massdriverDefinitionPattern = regexp.MustCompile(`^[a-zA-Z0-9]`)
...
 			}
-			hydratedSchema[key] = hydratedValue
 		}
-		return hydratedSchema, nil
+		return hydrateMap(ctx, c, cwd, hydratedSchema, schema)
 	default:
-		return any, nil
+		return anyVal, nil
+	}
+}
```
chrisghill, line 62: Honestly, it's pretty tough to track the changes in this file past this point. As long as the unit tests pass, we should be good.

pkg/provisioners/terraform/compile_test.go
chrisghill: Ha! Good catch. This is actually a better check. Terraform rejects an empty json file for tfvars doc:

```
$ terraform plan -var-file lol.tfvars.json
╷
│ Error: Missing value
│
│   on lol.tfvars.json line 1:
│    1:
│
│ The JSON data ends prematurely.
╵
╷
│ Error: Root value must be object
│
│   on lol.tfvars.json line 1:
│    1:
│
│ The root value in a JSON-based configuration must be either a JSON object or a JSON array of objects.
```

chrisghill (approved):
This is great! Glad we are cleaning this code up a bit.

Left some comments, all of them pretty minor so I'm gonna approve so you aren't blocked needing another review after addressing them.

cmd/definition.go
```diff
@@ -109,11 +111,12 @@ func runDefinitionPublish(cmd *cobra.Command, args []string) error {
 	byteValue, _ := ioutil.ReadAll(defFile)
 	var art definition.Definition
 
-	json.Unmarshal([]byte(byteValue), &art)
+	if jsonErr := json.Unmarshal(byteValue, &art); err != nil {
```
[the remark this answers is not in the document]
WillBeebe: I had the same question. From recent go-ing it's possible it's a different type than the earlier err variable? Curious what Jacob has to say.

pkg/application/convert_test.go
WillBeebe: Thanks for calling this out. I haven't communicated this until now but since this PR is pretty big and mostly style + improved go-isms I support getting this merged, then I can get my code up to date.

pkg/common/constants.go
WillBeebe: Plus one for package constants.

cmd/definition.go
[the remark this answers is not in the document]
jaketf: not shadowing the `err` var has some benefits:
- makes it more obvious it's scope
- makes it easier to move code around / remove sections without having to figure out "wait where do i need to add / remove this `:=` vs. `=`
- avoids situations like this where you wrap the wrong error or the error "falls through"
```go
a, err := foo()
// didn't catch err != nil
for i := range(list) {
  if err != nil {
    return fmt.Errorf("error processing list element %s %w", i-1, err)
  }
  err = bar(list[i])
}
```
if instead we had:
```go
```go
a, fooErr := foo()
// didn't catch err != nil
for i := range(list) {
  if barErr != nil {
    return fmt.Errorf("error processing list element %s %w", i-1, fooErr)
  }
  barErr = bar(list[i])
}
```
it'd be a lot easier for the reviewer to see "hey you're wrapping the wrong error here, also probably best to move that error handling beneath the `bar()` function call"

these examples are a bit simplistic but sort of illustrate the sentiement behind not shaddowing.

I think it's a little up for debate TBH and i see different go code bases decide to do different things. I personally find it a little more sane to think about his way

.github/workflows/test.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.17
+          go-version: 1.18
```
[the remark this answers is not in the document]
jaketf: good catch, definitely not! I'll revert to 1.17 we should upgrade all at once in a separate PR when necessary.

cmd/definition.go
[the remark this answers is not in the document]
chrisghill: Sounds good to me. Thanks for the explanation!

pkg/application/testdata/bundle/schema-params.json
```diff
@@ -1 +1 @@
-{"$id":"https://schemas.massdriver.cloud/schemas/bundles/my-app/schema-params.json","$schema":"http://json-schema.org/draft-07/schema","description":"An application","name":"my-app","properties":{"age":{"title":"Age","type":"integer"},"name":{"title":"Name","type":"string"}},"required":["name"]}
\ No newline at end of file
+{"$id":"https://schemas.massdriver.cloud/schemas/bundles/my-app/schema-params.json","$schema":"http://json-schema.org/draft-07/schema","description":"An application","name":"my-app","properties":{"age":{"title":"Age","type":"integer"},"name":{"title":"Name","type":"string"}},"required":["name"]}
```
WillBeebe, line 1: why is this one not pretty-printed?

pkg/bundle/publish.go
```diff
@@ -90,7 +90,8 @@ func (b *Bundle) generateBundlePublishBody() (BundlePublishPost, error) {
 }
 
 func UploadToPresignedS3URL(url string, object io.Reader) error {
-	req, err := http.NewRequest("PUT", url, object)
+	// TODO: is there a better place to pull this context from?
```
WillBeebe, line 93: 👌

pkg/bundle/schema.go
```diff
@@ -97,12 +97,12 @@ func GenerateSchema(schema map[string]interface{}, metadata map[string]string, b
 	var err error
 	var mergedSchema = mergeMaps(schema, metadata)
 
-	json, err := json.Marshal(mergedSchema)
+	json, err := json.MarshalIndent(mergedSchema, "", "    ")
```
WillBeebe, line 100: this is nice

WillBeebe (approved):
Read and learned a lot, thanks Jake!

pkg/application/generate.go
jaketf: 1. `gosec` doesn't want any file perms > 600.
  - I think the permissions we chose are intentional / not a security threat as we're not writing credentials / sensitive info.
2. `gomnd` doesn't want magic numbers. 
  - we could name these constants like `UNIX_PERM_OWNER_WRITE_EVERYONE_READ` which would make this more readable, but didn't seem totally worth it to me. happy to hear others' thoughts!

pkg/bundle/hydrate.go
jaketf: good question, it doesn't like that we don't catch the error if the interface cannot be cast to this type (it will actually trigger a panic).
I suppose the best practice thing to do here is check the type assertion ([example](https://go.dev/tour/methods/15))
```go
b.Artifacts, ok := hydratedArtifacts.(map[string]interface{})
if !ok {
  return errors.New("hydrated artifact could not be cast to map[string]interface")
}
```

I'll find all these and flip them to that pattern.

pkg/common/constants.go
jaketf: great suggestion!

you happen to know if the `fs` package has constants for common permissions like `0777` or `0644` (re: another comment on magic numbers?

pkg/application/testdata/bundle/schema-params.json
jaketf: good question! I'll fix that!

pkg/common/constants.go
WillBeebe: Good question. I just checked the `fs` package and didn't see any permission constants.

pkg/application/generate.go
chrisghill: We could probably get away with `0600` but that's pretty strict.

For the magic numbers, I'd rather do something[ like this ](https://stackoverflow.com/a/42718395)if we don't want to do `0644`:
That way we can additively `OR` the permissions we want, as opposed to having constants for each possible setting (of which there are 512 different combinations: 8 * 8 * 8)

pkg/common/constants.go
chrisghill: It doesn't, but I found this comment which I like: https://stackoverflow.com/a/42718395

pkg/jsonschema/hydrate.go
jaketf: this is really just moving existing logic into smaller functions, the hope was to make the code over all more readable (make the complexity linters happy) 😅

pkg/application/testdata/bundle/schema-params.json
jaketf: added `pretty-format-json` to pre-commit so we never have to remember again :)

pkg/application/generate.go
jaketf: > That assumes there is sensitive data in every file it seems. In this case there isn't.

yeah this is why i disabled that linter for this line. I think we agree this permission is necessary for our needs.

> like this

nice find, I'll add that!

pkg/jsonschema/hydrate.go
chrisghill: Yeah, I think its a good change 👍

PR #46, opened 2022-07-08
provide more info for 404 of artifacts

WillBeebe:
See the comments below for the yamls I used to test after writing tests for these. 

## Better-errors

I checked:

- massdriver ref
  - added code, added test
- http ref
  - added code, added test 
- local ref
  - the error was already informative, so I didn't add anything. 

## CLI Interface - Usage

I didn't check these changes in, but when an error is returned, CLI usage is printed out. I haven't seen this before in a CLI for an error, just for improper usage. When you type `mass bun build` then a helpful message is displayed. I think it's a better experience to not show usage when there's an internal CLI error.

*existing*
```
mass bundle build
{"level":"error","error":"open /Users/wbeebe/repos/massdriver-cloud/_test-hydrate/types/my-local-yolo-type.json: no such file or directory","time":"2022-07-09T15:10:31-07:00","message":"an error occurred while hydrating bundle"}
Usage:
  mass bundle build [flags]

Flags:
  -h, --help            help for build
  -o, --output string   Path to output directory (default is massdriver.yaml directory)

Global Flags:
  -k, --api-key string   Massdriver API key (can also be set via MASSDRIVER_API_KEY environment variable)
  -v, --verbose          Enable verbose output
```

*with* `SilenceUsage:  true,`
```
mass bundle build
{"level":"error","error":"open /Users/wbeebe/repos/massdriver-cloud/_test-hydrate/types/my-local-yolo-type.json: no such file or directory","time":"2022-07-09T15:10:31-07:00","message":"an error occurred while hydrating bundle"}
```

When a user is not using the CLI properly. I think this is good!
```
mass bun build
Error: unknown command "bun" for "mass"

Did you mean this?
	bundle

Run 'mass --help' for usage.
```

## CLI Interface - Errors

Right now, if an error is returned, it is printed twice. 

*existing*
```
{"level":"error","error":"open /Users/wbeebe/repos/massdriver-cloud/_test-hydrate/types/my-local-yolo-type.json: no such file or directory","time":"2022-07-09T15:14:03-07:00","message":"an error occurred while hydrating bundle"}
Error: open /Users/wbeebe/repos/massdriver-cloud/_test-hydrate/types/my-local-yolo-type.json: no such file or directory
```

*with* `SilenceErrors: true,`

```
mass bundle build
{"level":"error","error":"open /Users/wbeebe/repos/massdriver-cloud/_test-hydrate/types/my-local-yolo-type.json: no such file or directory","time":"2022-07-09T15:10:31-07:00","message":"an error occurred while hydrating bundle"}
```

I think 

+ twice isn't good
+ json is easier to parse when the cli is running in CI/CD w/o a human
+ non-json is more human-CLI friendly
  + to support this, we wouldn't use `SilenceErrors: true` we'd _not_ print the error at the top level of Cobra. We'd lean on the `loggingSetup` printing _I'm pretty sure_
Thoughts?

### DIFF
--- a/pkg/definition/get.go
+++ b/pkg/definition/get.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"path"
 
 	"github.com/massdriver-cloud/massdriver-cli/pkg/client"
@@ -31,7 +32,7 @@ func GetDefinition(c *client.MassdriverClient, definitionType string) (map[strin
 		return definition, err
 	}
 
-	if resp.StatusCode != http.StatusOk {
+	if resp.StatusCode != http.StatusOK {
 		fmt.Println(string(respBodyBytes))
 		return definition, errors.New("received non-200 response from Massdriver: " + resp.Status + " " + definitionType)
 	}
--- a/pkg/definition/get.go
+++ b/pkg/definition/get.go
@@ -33,7 +33,7 @@ func GetDefinition(c *client.MassdriverClient, definitionType string) (map[strin
 
 	if resp.Status != "200 OK" {
 		fmt.Println(string(respBodyBytes))
-		return definition, errors.New("received non-200 response from Massdriver: " + resp.Status)
+		return definition, errors.New("received non-200 response from Massdriver: " + resp.Status + " " + definitionType)
 	}
 
 	err = json.Unmarshal(respBodyBytes, &definition)
--- a/pkg/definition/get.go
+++ b/pkg/definition/get.go
@@ -31,7 +31,7 @@ func GetDefinition(c *client.MassdriverClient, definitionType string) (map[strin
 		return definition, err
 	}
 
-	if resp.Status != "200 OK" {
+	if resp.StatusCode != http.StatusOk {
 		fmt.Println(string(respBodyBytes))
 		return definition, errors.New("received non-200 response from Massdriver: " + resp.Status + " " + definitionType)
 	}
--- a/pkg/definition/get.go
+++ b/pkg/definition/get.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 	"path"
@@ -33,7 +32,6 @@ func GetDefinition(c *client.MassdriverClient, definitionType string) (map[strin
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		fmt.Println(string(respBodyBytes))
 		return definition, errors.New("received non-200 response from Massdriver: " + resp.Status + " " + definitionType)
 	}
 
--- a/pkg/jsonschema/hydrate_test.go
+++ b/pkg/jsonschema/hydrate_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/massdriver-cloud/massdriver-cli/pkg/client"
@@ -13,9 +14,10 @@ import (
 )
 
 type TestCase struct {
-	Name     string
-	Input    interface{}
-	Expected interface{}
+	Name                string
+	Input               interface{}
+	Expected            interface{}
+	ExpectedErrorSuffix string
 }
 
 func TestHydrate(t *testing.T) {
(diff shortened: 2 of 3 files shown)


pkg/definition/get.go
```diff
@@ -33,7 +33,7 @@ func GetDefinition(c *client.MassdriverClient, definitionType string) (map[strin
 
 	if resp.Status != "200 OK" {
```
chrisghill, line 34:
```suggestion
	if resp.StatusCode != http.StatusOk {
```

Let's get this updated while you're in there.

WillBeebe: Ooo great point. I'll comb for any others as well

chrisghill (approved):

WillBeebe:
Did some more paired-down testing.

## Massdriver ref not found

I added a test for this, but here's a local repro.

```yaml
connections:
  properties:
    aws_authentication:
      $ref: massdriver/aws-iam-who
```

```
{"level":"error","error":"received non-200 response from Massdriver: 404 Not Found massdriver/aws-iam-who","time":"2022-07-09T14:54:26-07:00","message":"an error occurred while hydrating bundle"}
```

## HTTP Ref not found 

I added a test for this, but here's a local repro.

```yaml
params:
  properties:
    moon_region:
      $ref: https://raw.githubusercontent.com/massdriver-cloud/artifact-definitions/main/definitions/types/moon-region.json
```

```
{"level":"error","error":"received non-200 response getting ref 404 Not Found https://raw.githubusercontent.com/massdriver-cloud/artifact-definitions/main/definitions/types/moon-region.json","time":"2022-07-09T14:56:01-07:00","message":"an error occurred while hydrating bundle"}
```

## Local ref not found

I didn't add any code for this, it already reported a useful error!

```yaml
params:
  properties:
    local_ref:
      $ref: ./types/my-local-yolo-type.json
```

```
{"level":"error","error":"open /Users/wbeebe/repos/massdriver-cloud/_test-hydrate/types/my-local-yolo-type.json: no such file or directory","time":"2022-07-09T14:57:10-07:00","message":"an error occurred while hydrating bundle"}
```

pkg/jsonschema/hydrate_test.go
```diff
@@ -6,16 +6,18 @@ import (
...
 )
 
 type TestCase struct {
-	Name     string
-	Input    interface{}
-	Expected interface{}
+	Name                string
+	Input               interface{}
+	Expected            interface{}
+	ExpectedErrorSuffix string
```
WillBeebe, line 20: Really open to feedback here, just hacked together something where I could test the errors.

chrisghill:
> twice isn't good

Agreed. We should suppress the usage. Its only needed if they didn't put in a valid command, not if the command execution failed.

> json is easier to parse when the cli is running in CI/CD w/o a human

Agreed. The CLI was copy-pasta'd from `xo`, which is why all the logging was still in JSON. Cory and I have discussed this and we should rip out the JSON logging and have readable outputs for users in the CLI

pkg/jsonschema/hydrate_test.go
chrisghill: @jaketf fixed up `cola` to do proper error testing.

Basically you can define error types: https://github.com/massdriver-cloud/cola/blob/bc538e01e188b002a7cdbc4c662fe5e071486f92/pkg/cidr/errors.go#L8-L9

Then when you error, you can specify to use that type:
https://github.com/massdriver-cloud/cola/blob/bc538e01e188b002a7cdbc4c662fe5e071486f92/pkg/cidr/errors.go#L8-L9

Then when testing, you can make sure the error matches that type:
https://github.com/massdriver-cloud/cola/blob/bc538e01e188b002a7cdbc4c662fe5e071486f92/pkg/cidr/errors.go#L8-L9
https://github.com/massdriver-cloud/cola/blob/bc538e01e188b002a7cdbc4c662fe5e071486f92/pkg/cidr/errors.go#L8-L9

WillBeebe:
> > twice isn't good
> 
> Agreed. We should suppress the usage. Its only needed if they didn't put in a valid command, not if the command execution failed.
> 
> > json is easier to parse when the cli is running in CI/CD w/o a human
> 
> Agreed. The CLI was copy-pasta'd from `xo`, which is why all the logging was still in JSON. Cory and I have discussed this and we should rip out the JSON logging and have readable outputs for users in the CLI

I'm happy to do that. I can create a ticket for _just_ that, separate from this PR.

pkg/jsonschema/hydrate_test.go
WillBeebe: That looks pretty slick and I like it. I leaned towards suffix / prefix here because the url was reported in the error and the port changes every time. I thought about regex and maybe I can do the Cola pattern but w/ Regex. I'll take a look.

WillBeebe: Made a ticket, will come back to:
https://github.com/massdriver-cloud/massdriver/issues/1124

PR #54, opened 2022-07-17
JSON Schema name -> title

coryodaniel:
Top-level field name in JSON Schema is "title" not "name".


WillBeebe (approved):

coryodaniel:
Note, this is broken still. I'm not sure how to test this correctly. @chrisghill did you get a chance to pull this?

chrisghill:
@coryodaniel Fixed this in the last commit.

The package test stuff is pretty fragile right now. We generate a bundle, then compare the generated bundle against a "test folder". Since we don't have a good way to compare the 2 directories for equality, we just do an MD5 hash compare. This breaks if the json fields get ordered differently in the generated file.

It'd be great to eventually do a logical comparison, but this is fixed for now.

PR #62, opened 2022-07-23
adding files to bundle template, update massdriver.yaml

WillBeebe:
### Added
- pkg/bundle/templates/terraform/Makefile
- pkg/bundle/templates/terraform/tf_helper.rb

### Bugfix
- bundle template was missing steps in `massdriver.yaml`


pkg/bundle/templates/terraform/massdriver.yaml
```diff
@@ -6,6 +6,10 @@ ref: github.com/YOUR_NAME_HERE/{{ .Name }}
 access: "{{ .Access }}"
 type: "{{ .Type }}"
 
+steps:
```
coryodaniel, line 9: Lets check in to see if @chrisghill got the fix into xo for this. I'd like not to expose this functionality to people. Its something for them to mess up with no feature to gain.

pkg/bundle/templates/terraform/tf_helper.rb
```diff
@@ -0,0 +1,58 @@
+#! ruby
```
coryodaniel, line 1: I think this is fine to add but it'll be getting replaced fairly soon. I was planning to convert this to go using terratest so we could start automating the tf tests we are doing locally in CI.

coryodaniel (approved):
Check w/ chris re: steps. I know it makes all of our bundles broken so we want to add it, but fixing it in xo is the better choice if it can happen quickly.

See notes re ruby script.

pkg/bundle/templates/terraform/massdriver.yaml
WillBeebe: I'll yank it from the template

PR #65, opened 2022-07-25
Remove stories.js from bundle template

WillBeebe:
verified w/ the frontend team that they don't need that w/ the new bundle dev approach


WillBeebe:
Talking w/ burger we won't need this file in the bundle template

coryodaniel (approved):

PR #66, opened 2022-07-27
clone app templates

WillBeebe:
This is about as small as I can make the app cloning functionality. It doesn't touch any existing methods either 👍 

resolves: https://github.com/massdriver-cloud/massdriver/issues/1175


cmd/application.go
```diff
@@ -26,10 +29,17 @@ var applicationCmd = &cobra.Command{
 var applicationGenerateCmd = &cobra.Command{
 	Use:     "generate",
 	Aliases: []string{"gen"},
-	Short:   "Generates a new application template",
+	Short:   "Deprecated: Generates a new application template",
 	RunE:    runApplicationGenerate,
 }
 
+var applicationNewCmd = &cobra.Command{
+	Use:     "new",
```
coryodaniel, line 37: <3

cmd/application.go
```diff
@@ -38,11 +48,27 @@ var applicationPublishCmd = &cobra.Command{
 	SilenceUsage: true,
 }
 
+var applicationTemplatesCmd = &cobra.Command{
+	Use:     "templates",
```
coryodaniel, line 52: This is great for now (and no need to change based on my comment) ->

I could see us add a 'repo' subcommand `mass repo` for managing bundle and app repos.

cmd/application.go
```diff
@@ -38,11 +48,27 @@ var applicationPublishCmd = &cobra.Command{
...
+	Aliases: []string{"plates"},
```
coryodaniel, line 53: Typo?

coryodaniel, line 53:
```suggestion
	Aliases: []string{"tmpl"},
```

pkg/application/new.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	errCopy := copyTemplate(source, data.TemplateName, data.OutputDir)
+	if errCopy != nil {
+		log.Err(errCopy).Msg("error copying template")
+		return errCopy
+	}
+
+	// TODO: only do this for templates w/ a helm chart
+	if errModifyHelm := modifyHelmTemplate(*data); errModifyHelm != nil {
```
coryodaniel, line 43: Should we even differentiate? We can interpolate straight into _any_ file using go templates. We'd need to change the delim to support helm since its the same.

pkg/application/new.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+
+		return nil
+	})
+}
+
+// TODO: both of these modify methods will use golang templating in the future
+func modifyAppYaml(data TemplateData) error {
+	appYAML, _ := Parse(data.OutputDir + "/app.yaml")
+	// TODO: Cory has a PR to change this to title
```
coryodaniel, line 86: That PR is to only fix invalid JSON Schema. Our field is `name` json schema calls it `title`

pkg/application/new.go
```diff
@@ -0,0 +1,123 @@
...
+
+	errWrite := ioutil.WriteFile(path.Join(data.OutputDir, "app.yaml"), appYAMLBytes, common.AllRead|common.UserRW)
+	if errWrite != nil {
+		return errWrite
+	}
+
+	return nil
+}
+
+func modifyHelmTemplate(data TemplateData) error {
```
coryodaniel, line 104: I dont think we should have any special treatment in `mass app new` ... we should standardize on a templating language for the templates to be in and people have to learn that [liquid](https://shopify.github.io/liquid/) is a good cross language template that many app devs are familiar with.

If we go down the 'we do special shit for helm' we'll need to do special processing for `kustomize`, the `serverless framework`, `sam`, etc... We should treat the templates dirs as a bag of interpolatable strings.

pkg/application/new.go
WillBeebe: I don't think we should. This was the quickest path to new command + old functionality. I can do the templating work in the next PR

chrisghill (approved):
Looking good so far for a first pass. A few notes I'm sure you're aware of:
* Let's make sure we get tests on stuff as it becomes more formalized
* I didn't review the cache stuff too much since I'm expecting that to change in the near future (to more of the helm package management)

Approving now to unblock you once you address the comments here.

pkg/application/new.go
coryodaniel: 🚀

cmd/application.go
```diff
@@ -38,11 +48,27 @@ var applicationPublishCmd = &cobra.Command{
...
+	Short:   "Lists available application templates",
+	RunE:    runApplicationTemplates,
+}
+
+var templatesRefreshCmd = &cobra.Command{
```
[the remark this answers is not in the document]
WillBeebe: I think you're right here. It sounds like some of these will live under `mass repo xyz`

pkg/application/new.go
```diff
@@ -0,0 +1,123 @@
...
+		if writeErr != nil {
+			return writeErr
+		}
+
+		return nil
+	})
+}
+
+// TODO: both of these modify methods will use golang templating in the future
+func modifyAppYaml(data TemplateData) error {
```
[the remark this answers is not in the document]
WillBeebe: Very helpful, thank you!

pkg/application/new.go
```diff
@@ -0,0 +1,123 @@
...
+
+			return os.Mkdir(outputPath, common.AllRWX)
+		}
+
+		file, readErr := os.ReadFile(templateDir + "/" + templateName + "/" + filePath)
+		if readErr != nil {
+			return readErr
+		}
+
+		writeErr := os.WriteFile(outputPath, file, common.AllRWX)
```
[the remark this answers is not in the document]
WillBeebe: Great catch, thank you

PR #69, opened 2022-07-29
mass version command and latest version check

jaketf:
- adds  a `version` subcommand to:
  - print the current version info 
  - check if there are any newer releases of the CLI and print a warning if so.
  - print the git SHA it was built from (sometimes useful incase for whatever reason we push to a mutable tag and are confused)
- modifies our releasing workflow to dynamically inject version and git SHA into our binary.
- adds a `-version` flag (some gophers might expect this to be here and it's easy to give them that behavior w/ cobra)

fixes https://github.com/massdriver-cloud/massdriver/issues/1190

## Notes for reviewer
- if you're not familiar `ldflags` is a way to set dynamic values at compile time often used for versioning when running `go build`
- I adjusted the make commands such that they'd set version and git SHA to indicate that this was built locally.

## Question for anyone who knows
- I'd love to do a "test run" of the release flow with the new LD flags before making it official could we do that with a pre-release in github?


jaketf:
truly honored, and humbled to have grand pleasure of being PR #69 on this repo ;)

coryodaniel:
I’m not positive about the pre-release stuff either. This all looks great like the use of LDflags.

I’m curious if we need the extra -version. It’s extra code, more output in help, and a second way to do something.

jaketf:
> I’m curious if we need the extra -version. It’s extra code, more output in help, and a second way to do something.

I don't have a strong opinion and am happy to remove.
FWIW Cobra does all the implementation so it's just a one line of setting an extra struct field so maintenance burden is minimal. I take your point on multiple ways to do things / help duplication.

My original thought was that it might save someone the churn of `-version flag unrecognized -> read help string -> ok i'll remove the dash`.
I've definitely more than once been like and wished it just was behind all of those:
```bash
$ python version # no dice
can't open file 'version': [Errno 2] No such file or directory
$ python -version # grrrr
Unknown option: -e
$ python --version # FINALLY
3.9.13
```

cmd/root.go
```diff
@@ -21,6 +22,9 @@ examples and usage of using your application. For example:
 Cobra is a CLI library for Go that empowers applications.
 This application is a tool to generate the needed files
 to quickly create a Cobra application.`,
+	// adds the canonical -verison flag to the root command which some users might expect
+	// note we are also adding mass version subcommand which will additionally print a warning if there is a newer version of the CLI
+	Version: version.MassVersion(),
```
jaketf, line 27: @coryodaniel feel free to commit this if you want to remove the alternate `-version` flag
```suggestion
```

coryodaniel: Seeing your comment above, what I would expect is `mass version` to work. 

I can’t tell what the command is from seeing the code here.

cmd/version.go
```diff
@@ -0,0 +1,35 @@
...
+	Long:    ``,
+	Run:     runVersion,
+}
+
+func runVersion(cmd *cobra.Command, args []string) {
+	isOld, latest, err := version.CheckForNewerVersionAvailable()
+	if err != nil {
+		fmt.Printf("could not check for newer versions at %v: %v. skipping...\n", version.LatestReleaseURL, err.Error())
+	} else if isOld {
+		fmt.Printf("WARNING: there is a new version of the mass CLI %v you can download the appropriate binary from %v\n", latest, version.LatestReleaseURL)
```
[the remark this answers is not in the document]
WillBeebe: Thoughts on softening the message and not printing the latest version? I was thinking the latest lacked context of current version.

cmd/root.go
```diff
@@ -21,6 +22,8 @@ examples and usage of using your application. For example:
 Cobra is a CLI library for Go that empowers applications.
 This application is a tool to generate the needed files
 to quickly create a Cobra application.`,
+	// adds the canonical -verison flag to the root command which some users might expect
+	// note we are also adding mass version subcommand which will additionally print a warning if there is a newer version of the CLI
```
WillBeebe, line 26:
```suggestion
```

WillBeebe (approved):
Good work Jake, LGTM 🍾

cmd/version.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
```
coryodaniel, line 2:
```suggestion
```

coryodaniel (approved):
Looks good lets remove the boilerplate comments in suggestions.

We should gut the rest of the cobra boilerplate comments as well (not needed for this PR just a note)

PR #70, opened 2022-07-29
test release workflow on pre-release

jaketf:
from: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release
```
Note: The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.
```


WillBeebe (approved):
Approved to test

PR #74, opened 2022-07-31
template tests

WillBeebe:
- os filesystem templates (app-templates) have tests
- files embedded in the Go binary (bundle-templates) have tests
- both use golangs `text/template`

resolves: https://github.com/massdriver-cloud/massdriver/issues/1207


pkg/application/prompt.go
```diff
@@ -21,14 +22,15 @@ var prompts = []func(t *TemplateData) error{
 	getLocation,
 }
 
-var promptsNew = []func(t *TemplateData) error{
+var promptsNew = []func(t *template.Data) error{
 	getName,
 	getDescription,
 	getAccessLevel,
 	getTemplate,
+	getOutputDir,
```
WillBeebe, line 30: prompt for output dir, like bundles

pkg/template/files_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"io/ioutil"
+	"path"
+	"testing"
+
+	gotmpl "text/template"
+
+	"github.com/massdriver-cloud/massdriver-cli/pkg/template"
+)
+
+func TestWriteTemplateToFile(t *testing.T) {
```
xpositivityx, line 14: At some point we should figure out how to use the IO write to write this to a buffer so we can run tests without access to a filesystem.

xpositivityx (approved):

pkg/template/files_test.go
WillBeebe: Thanks for the reminder on that one. 👍 I'll add a ticket to the backlog to track.

PR #78, opened 2022-08-03
app build / package / publish

WillBeebe:
This gets us

`mass app build` so we can run Terraform locally like we do with bundles.


pkg/application/prompt.go
```diff
@@ -149,6 +149,20 @@ func getTemplate(t *template.Data) error {
 		return err
 	}
 
+	if strings.HasPrefix(result, "kubernetes-") {
```
WillBeebe, line 152: cheat code

If it's kubernetes prefixed, ask the cloud provider. This is used for conditionals in the template

pkg/application/parse_test.go
```diff
@@ -16,16 +16,13 @@ func TestParse(t *testing.T) {
...
 			name:    "simple",
-			appPath: "./testdata/appsimple.yaml",
+			appPath: "./testdata/appsimple/massdriver.yaml",
 			want: application.Application{
 				Schema:      "draft-07",
 				Name:        "my-app",
 				Description: "An application",
 				Ref:         "github.com/user/app",
 				Access:      "private",
-				Deployment: application.Deployment{
```
WillBeebe, line 26: Not quite removed from the `Application` object yet, but we're close!

pkg/application/package_test.go
```diff
@@ -64,10 +64,7 @@ func TestPackage(t *testing.T) {
 			// Create a temp dir, write out the archive, then shell out to untar
 			testDir := t.TempDir()
 
-			application.SimpleParams = `{"properties":{"simple":{"type":"string"}}}`
```
WillBeebe, line 67: no more "default params" / "simple app"

pkg/application/package.go
```diff
@@ -116,21 +103,3 @@ func generateStep(step bundle.Step, workingDir, bundlePath string) error {
 	}
 	return nil
 }
-
-func packageChart(chartPath string, destPath string) error {
```
WillBeebe, line 120: no longer need this. This DOES assume the chart is in `src` because we copy all the contents of `src` (and other steps) when packaging. That name of the `chart` directory can be anything.

```
massdriver.yaml
src/
  chart/  
```

pkg/application/package.go
```diff
@@ -85,8 +59,21 @@ func PackageApplication(appPath string, c *client.MassdriverClient, workingDir s
 	if err != nil {
 		return nil, err
 	}
+	appDir := filepath.Dir(appPath)
+	bundleDir := filepath.Dir(bundlePath)
+	// Make all directories, generate provisioner-specific files
+	for _, step := range steps {
+		err = os.MkdirAll(path.Join(workingDir, step.Path), 0744)
```
WillBeebe, line 66: the `generateStep` function will actually make the dir but this felt better

- create output dir
- copy from source / to output dir
- generate files to output dir (generate files wins)

cmd/build.go
```diff
@@ -0,0 +1,50 @@
...
+	Use:  "build",
+	Long: ``,
+	RunE: runBuild,
+}
+
+func init() {
+	rootCmd.AddCommand(buildCmd)
+}
+
+func runBuild(cmd *cobra.Command, args []string) error {
```
WillBeebe, line 21: an experiment, what do you think?

`mass build`

cmd/bundle.go
```diff
@@ -60,62 +58,25 @@ func init() {
...
-			err = terraform.GenerateFiles(output, step.Path)
-			if err != nil {
-				log.Error().Err(err).Str("provisioner", step.Provisioner).Msg("an error occurred while generating provisioner files")
-				return err
-			}
-		default:
-			log.Error().Msg("unknown provisioner: " + step.Provisioner)
-			return fmt.Errorf("unknown provisioner: %v", step.Provisioner)
-		}
+	bun, err := bundle.Parse(configFile, nil)
```
coryodaniel, line 76: Does this not throw a warning for not using `err`?

cmd/definition.go
```diff
@@ -82,19 +75,14 @@ func runDefinitionGet(cmd *cobra.Command, args []string) error {
...
 		return err
 	}
 
-	apiKey, err := cmd.Flags().GetString("api-key")
-	if err != nil {
-		return err
-	}
-	if apiKey != "" {
-		c.WithAPIKey(apiKey)
+	c, errClient := initClient(cmd)
```
coryodaniel, line 83: dig it.

pkg/application/convert.go
```diff
@@ -73,29 +34,13 @@ func (app *Application) ConvertToBundle() (*bundle.Bundle, error) {
...
-
 	b.Artifacts["properties"] = make(map[string]interface{})
 
-	// UI
-	if app.Deployment.Type == "simple" {
-		b.UI = make(map[string]interface{})
-		if jsonErr := json.Unmarshal([]byte(SimpleUI), &b.UI); jsonErr != nil {
-			return b, fmt.Errorf("error parsing simple ui: %w", jsonErr)
-		}
+	if app.UI != nil {
```
coryodaniel, line 39: tight

pkg/application/package.go
```diff
@@ -17,65 +15,43 @@ import (
...
-	appYaml, err := os.Create(path.Join(workingDir, "app.yaml"))
-	if err != nil {
-		return nil, err
-	}
-	defer appYaml.Close()
-	appYamlBytes, marshalErr := yaml.Marshal(*app)
-	if marshalErr != nil {
-		return nil, marshalErr
+	bytes, errMarshal := yaml.Marshal(*app)
+	errWrite := common.WriteFile(path.Join(workingDir, "app.yaml"), bytes, errMarshal)
```
coryodaniel, line 26: how does this work w/ `app.yaml`? is this a temp file for the build process?

pkg/application/package.go
```diff
@@ -17,65 +15,43 @@ import (
...
-		if err != nil {
-			return nil, err
-		}
-		err = packageChart(path.Join(path.Dir(appPath), app.Deployment.Path), path.Join(workingDir, "chart"))
-		if err != nil {
-			return nil, err
+	// if the bundle doesn't have Steps
+	// use default steps to generate the provisioner files
+	steps := b.Steps
+	if b.Steps == nil {
```
coryodaniel, line 46: This might be duplicated in bundle packages' code. Not sure if its worth drying up.

pkg/application/package.go
coryodaniel: In bundle generate (we should probably rename to `new` as well) we now check if the file exists and prompt for overwrite so if the template gets updated w/ a new file you can run the command again and get the new files. Might be nice here especially if we are rapidly riffing through an app template.

pkg/application/parse_test.go
```diff
@@ -16,16 +16,13 @@ func TestParse(t *testing.T) {
 	tests := []test{
 		{
 			name:    "simple",
-			appPath: "./testdata/appsimple.yaml",
+			appPath: "./testdata/appsimple/massdriver.yaml",
 			want: application.Application{
 				Schema:      "draft-07",
 				Name:        "my-app",
 				Description: "An application",
 				Ref:         "github.com/user/app",
```
coryodaniel, line 24: An aside, we should rename `ref` to `source` (this will hit all bundles and code in md & maybe xo). This is going to be real confusing vs `$ref` in a JSON Schema centric codebase. Lets get a ticket open to coordinate that.

We don't actually do any logic on it today, but things we can do with it in the future:

1. Linking to the source from the bundle card in the drawer (or a bundle modal) so devs and auditors can easily get to code
2. When displaying IaC errors in the new provisioning errors UI we could actually deep link to the file (tf sends the file name & line num) if we had an ops-type person that wanted to submit a PR.

pkg/application/prompt.go
coryodaniel, line 152: Lets nuke this based on convo. `_providers.tf` will be a nightmare for a few more days, but this could be really confusing to someone trying to make a bundle that works on any cloud.

pkg/application/testdata/appcustom/massdriver.yaml
```diff
@@ -4,12 +4,14 @@ description: An application
...
-  type: simple
-
 params:
   required: ["name"]
   properties:
     name:
       type: string
       title: Name
+
+dependencies:
```
coryodaniel, line 14: So `dependencies` here like the literal structure _isnt_ JSON Schema (no `properties` key underneath it). We are immediately doing a transform on it and we have a lot of fields that aren't json schemy, but I could see a user being confused by that. We may want to highlight that dependencies isn't json schema in our App guide on docs.massdriver.cloud.

coryodaniel (commented):
Looking pretty good. Did you yank the experimental thing, I went through twice and didn't find it.

cmd/bundle.go
chrisghill: Probably cuz it gets used on line 82. But good catch, we should handle this error here.

pkg/application/package.go
chrisghill: The original k8s application modules were setup to use `app.yaml` instead of `massdriver.yaml`.  
https://github.com/massdriver-cloud/massdriver-bundles/blob/main/provisioners/terraform/modules/k8s-application/locals.tf#L13

We should swap this over to `massdriver.yaml`. Doesn't have to be this PR if you want to synchronize it w/ changing the module, but we should def change over.

pkg/application/package.go
chrisghill: Technically I don't think we even need this here. I put this in the bundle code to "default" the steps, but it ended up being useless since we only manipulate it in memory...we don't write the default steps out to disk (cuz we wanted the field hidden). When we package the bundle, we ship the file on disk...so it didn't get added. Hence why the default step stuff was broken for a bit, and Beebe fixed it with adding the default step to the provisioner script itself: https://github.com/massdriver-cloud/massdriver-bundles/blob/main/provisioners/terraform/get-steps.sh#L6-L8

I think this `bundle.yaml` file is ephemeral - we only need it so we can re-use the bundle package endpoint and logic to publish applications. @WillBeebe its up to you if you want to do this different for bundles and apps (default the steps for bundles in the provisioner script, but default the steps for apps in the bundle.yaml). I'm leaning towards consistency. What do you think?

pkg/application/package.go
```diff
@@ -85,8 +61,21 @@ func PackageApplication(appPath string, c *client.MassdriverClient, workingDir s
 	if err != nil {
 		return nil, err
 	}
+	appDir := filepath.Dir(appPath)
+	bundleDir := filepath.Dir(bundlePath)
```
chrisghill, line 65: Is the application and bundle in different directories? This is bit confusing since applications used to just be different types of bundles (but the same directory).
Edit: Oh, is the "bundlePath" the temporary directory where we're building the bundle and the appPath would be where the `massdriver.yaml` file is and the command was executed from?

pkg/application/package.go
chrisghill: I think this is okay for now.  I can envision a company that wants their helm chart at the top level, not nested in our directories, but we can figure that out if/when we get there.

pkg/application/package_test.go
chrisghill: Love it!

pkg/bundle/build.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/massdriver-cloud/massdriver-cli/pkg/client"
+	"github.com/massdriver-cloud/massdriver-cli/pkg/provisioners/terraform"
+
+	"github.com/rs/zerolog/log"
+)
+
+const configFile = "massdriver.yaml"
+
+func (b *Bundle) Build(c *client.MassdriverClient, output string) error {
```
chrisghill, line 14: Great change moving this into a function. Meant to do that for a while.

pkg/application/prompt.go
chrisghill: Like Cory mentioned, I think we can figure out a way to keep kubernetes cloud agnostic and "magical" without asking them about the cloud.

chrisghill (commented):
A lot of really great changes. Only change I saw is removing that kubernetes cloud prompt

pkg/application/package.go
WillBeebe: I haven't tried to run another "gen / new" in the same directory but I kept this in mind. The same `template` pkg bundles and app have this for the writing _file_. https://github.com/massdriver-cloud/massdriver-cli/blob/main/pkg/template/files.go#L70-L78

I think that will error out on an existing directory but the guts are there for run again and ask before override.

pkg/application/package.go
WillBeebe: I will come back to this, I'll update the module code at the same time

pkg/application/package.go
WillBeebe: I'll grab this as part of the "unify building apps and bundles"

pkg/application/parse_test.go
WillBeebe: I like it! I'll plop a ticket in the backlog for it

pkg/application/testdata/appcustom/massdriver.yaml
WillBeebe: Noted, can do

pkg/application/package.go
```diff
@@ -85,8 +61,31 @@ func PackageApplication(appPath string, c *client.MassdriverClient, workingDir s
...
+	// Make all directories, generate provisioner-specific files
+	for _, step := range steps {
+		err = os.MkdirAll(path.Join(workingDir, step.Path), 0744)
+		if err != nil {
+			return nil, err
+		}
+		// ignore these, copy the rest
+		ignores := []string{
```
WillBeebe, line 73: This is new for apps, since we copy the everything from the src directory (chart) we want to ignore some things.

coryodaniel:
Hold on merging this, we found something interesting hacking tonight. Need to sync in AM.

PR #79, opened 2022-08-03
App temp refresh

jaketf:
pre-commit to ensure application tempaltes used as test data are up to date with massdriver/application-templates
for context there were some things out of date including app.yaml and i wanted to sort this out so i could accurately be testing when adding app dependencies to the `mass application new` command.

most of this PR is just added files from `application-templates` into the testdata folder. I don't know if we want to have so much of that duplicated here but at least it's automated in a sane-ish way.


WillBeebe (approved):
Holy wow / you're a machine. :) 

Ideally we have a few simple and small use-cases for the test data. Since it's in flux, let's keep most of this. 

I'd say

- don't add it as a precommit
- add a Make command to do this

then we have update-able templates and will in the future will probably pair that down even more.

jaketf:
linter was [complaining](https://github.com/massdriver-cloud/massdriver-cli/runs/7684050223?check_suite_focus=true) about some issues unrelated to this PR trying to fix...

jaketf:
issue seems related to the latest release of [v1.48.0 golangci-lint last night](https://github.com/golangci/golangci-lint/releases/tag/v1.48.0) it mostly adds go 1.19 support which we are not using. pinning our linter version for now.

PR #83, opened 2022-08-05
prompt for dependency name

WillBeebe:
Screenshot 2022-08-05 at 9 14 46 PM
Screenshot 2022-08-05 at 9 14 52 PM
Screenshot 2022-08-05 at 9 15 02 PM
Screenshot 2022-08-05 at 9 15 27 PM

The regex for dependency names:

Screenshot 2022-08-05 at 5 42 21 PM


pkg/template/testdata/application-templates-want/kubernetes-cronjob/massdriver.yaml
```diff
@@ -64,7 +64,7 @@ dependencies:
     required: true
 
 
-  massdriver/draft-node_0:
+  massdriver_draft_node:
```
WillBeebe, line 67: I'll pair down the test data later.

WillBeebe:
I'm going to prompt for the name instead, leaving this open to track.

coryodaniel:
A prompt would be good so people don’t leave massdriver_ in forever. We should regex it to a-z0-9_

WillBeebe:
@coryodaniel prompt added! I landed on nearly the same regex. I don't allow it to end in an underscore is the only difference.

coryodaniel (approved):
Tight!

xpositivityx (approved):

PR #85, opened 2022-08-08
Adding app template test cases

coryodaniel:
When testing, we need to know _what_ failed in application templates, not if app templates elsewhere are rendering correctly or not.

* [added] two happy path test cases
  * `renders-successfully` is a basic happy path test of interpolation
  * `renders-dependencies` is a test on .Dependencies being interpolated
* [removed] copy/pasta from application-templates
* [added] liquid whitespace trimming (we'll need to do these in our templates)
* [added] diff output on difference (doesnt look great, best output was HTML, not positive how to get the diffs to show correctly when logged from `go test`


pkg/template/render_test.go
```diff
@@ -70,9 +70,60 @@ func TestAppFromTemplate(t *testing.T) {
...
 			if gotMD5 != wantMD5 {
-				// TODO: need to print out which file is different
 				t.Errorf("got %v, want %v", gotMD5, wantMD5)
+				walkAndCompare(tc.wantPath, templateData.OutputDir)
+			}
+		})
+	}
+}
+
+func walkAndCompare(wantDir string, gotDir string) {
```
xpositivityx, line 80: SIIIIIck

pkg/template/testdata/application-templates/renders-dependencies/massdriver.yaml
```diff
@@ -0,0 +1,28 @@
...
+    hostname:
+      type: string
+    app_env:
+      type: string
+
+dependencies:
+  database:
+    type: massdriver/postgresql-authentication
+    required: true
+<md- range $key, $art:= .Dependencies md>
```
WillBeebe, line 19: great test case, all we need

xpositivityx (approved):

pkg/template/render_test.go
WillBeebe, line 80: Thank you, this is much more helpful that the directory compare.

xpositivityx:
This is great. Good TDD means light tests with limited code bloat.

WillBeebe (approved):
LGTM, thanks for paying down debt

pkg/template/render_test.go
coryodaniel: The diff tool is pretty powerful, there’s just something about how go test handles std out that it loses the fancy formatting.

PR #86, opened 2022-08-09
update gitignore .vscode/

jaketf:
we've run into this dir getting accidentally committed in a few recent PRs.

### DIFF
--- a/pkg/bundle/templates/terraform/.gitignore
+++ b/pkg/bundle/templates/terraform/.gitignore
@@ -124,3 +124,7 @@ schema-ui.json
 src/_connections_variables.tf.json
 src/_md_variables.tf.json
 src/_params_variables.tf.json
+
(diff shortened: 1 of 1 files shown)


WillBeebe:
Some devs probably want to add this ya? I think it's the legacy `.devcontainer` folder, better abstractions w/ `.devcontainer` around VSCode options etc.

However, I'm fine to add this if it helps us / most people now. My only ask is, help us pay down some debt and move this section
https://github.com/massdriver-cloud/massdriver-cli/blob/main/pkg/bundle/templates/terraform/.gitignore#L48-L50

Down to where yours is. Everything above `# Massdriver generated files` is generated by a website and I'd like to avoid interweaving los dos.

WillBeebe (approved):
Thank you, LGTM

PR #87, opened 2022-08-11
App is_a bundle

WillBeebe:
- apps _are_ bundles
- apps have connections, they are JSON schema 
- no more dependencies
- updated templates to work w/ connection prompt
- updated tests


pkg/bundle/templates/terraform/massdriver.yaml
```diff
@@ -27,15 +27,12 @@ params:
 # schema-connections.json
 # Named list of input connections (artifacts) (map[name]artifact)
 connections:
```
WillBeebe, line 29: support no-connections

coryodaniel:
Example YAML, I approximated the names of things.

```yaml
schema: draft-07
name: express-app-demo
description: An example express application using Massdriver
ref: github.com/massdrover-cloud/express-app-demo
access: private
type: application

params:
  properties:
    name:
      type: string
    log_level:
      type: string
      enum: ["warn", "error", "info"]
    namespace:
      type: string
      default: default
    replication:
      type: string
      enum:
        - async
        - sync

connections:
  required: "*"
  properties:
    kubernetes_cluster:
      $ref: massdriver/k8s
    mongo:
      $ref: massdriver/mongo-authentication
    sqs:
      $ref: massdriver/aws-sqs-pubsub-subscription
    s3:
      $ref: massdriver/aws-s3-application-assets-bucket

app:
  envs:
    MONGO_USERNAME: connections.mongo.path.to.username
    LOG_LEVEL: params.log_level
    STRIPE_KEY: secrets.ecomm_site_stripe_key # these would be set in MD at the target level; TBD in future

  policies:
    - connections.sqs.path.to.read.policy
    - connections.s3.path.to.write.policy

```

pkg/bundle/templates/terraform/massdriver.yaml
coryodaniel: We always need a connection schema, but it can be blank. If we want to allow people to omit the connections field in YAML we can generate a {} schema in the conns file

coryodaniel: We always need a connection schema, but it can be blank. If we want to allow people to omit the connections field in YAML we can generate a {} schema in the conns file

WillBeebe:
I'm working on the no-connections default for templates but otherwise ready for review!

coryodaniel:
I’m going to look in AM!

Cory ODaniel
Mobile: 310.606.0257
***@***.***
medium: @coryodaniel ***@***.***>
twitter: @coryodaniel <https://twitter.com/coryodaniel>

WillBeebe:
Sounds good! I had a little revelation and app package is much simpler now.

cmd/bundle.go
```diff
@@ -31,10 +30,16 @@ var bundleBuildCmd = &cobra.Command{
 var bundleGenerateCmd = &cobra.Command{
 	Use:     "generate",
 	Aliases: []string{"gen"},
-	Short:   "Generates a new bundle",
+	Short:   "Deprecated: Generates a new bundle",
 	RunE:    runBundleGenerate,
 }
 
+var bundleNewCmd = &cobra.Command{
+	Use:   "new",
```
WillBeebe, line 38: calling out adding `mass bundle new`, it does the same as `generate`

pkg/application/types.go
```diff
@@ -3,219 +3,37 @@ package application
 import "github.com/massdriver-cloud/massdriver-cli/pkg/bundle"
 
 type Application struct {
-	Schema       string                  `json:"schema" yaml:"schema"`
-	Name         string                  `json:"name" yaml:"name"`
-	Description  string                  `json:"description" yaml:"description"`
-	Ref          string                  `json:"ref" yaml:"ref"`
-	Type         string                  `json:"type" yaml:"type"`
-	Access       string                  `json:"access" yaml:"access"`
-	Deployment   Deployment              `json:"deployment" yaml:"deployment"`
```
WillBeebe, line 12: yanked

- Deployment
- Dependencies

added

- Connections
- Artifacts
- App

cmd/bundle.go
```diff
@@ -69,17 +83,24 @@ func runBundleBuild(cmd *cobra.Command, args []string) error {
...
 	}
 	if output == "" {
-		output = filepath.Dir(configFile)
+		output = filepath.Dir(common.MassdriverYamlFilename)
 	}
 
 	log.Info().Msg("building bundle")
-	bun, err := bundle.Parse(configFile, nil)
-	if errBuild := bun.Build(c, output); errBuild != nil {
+	b, err := bundle.Parse(common.MassdriverYamlFilename, nil)
```
xpositivityx, line 90: I feel like this same chain of events happens with a couple different functions. I wonder if it would be worth adding checkIsBundle and checkIsApplication to the bundle and app types and just dependency injecting that in to a more general version of this code.

xpositivityx (commented):
overall there are some really great refactors in here!

pkg/application/types.go
coryodaniel: Nice artifacts will be blank here for a while, I'm pretty sure publish requires them, so it can be an empty map if not present in the yaml file.

pkg/application/types.go
```diff
@@ -3,219 +3,37 @@ package application
...
 }
 
-type Deployment struct {
-	Type       string `json:"type" yaml:"type"`
-	Path       string `json:"path,omitempty" yaml:"path,omitempty"`
-	Chart      string `json:"chart,omitempty" yaml:"chart,omitempty"`
-	Repository string `json:"repository,omitempty" yaml:"repository,omitempty"`
+type AppBlock struct {
+	Envs     map[string]string `json:"envs" yaml:"envs"`
+	Policies []string          `json:"policies" yaml:"policies"`
```
coryodaniel, line 22: I think policies should be an array, but tell me if there is any value in them being a map ... with the way we process them in `for_each` in tf across all the clouds its just iterating on the array. Not sure if there is any value to having a named reference to them. Probably not. :bike: :shed:

pkg/application/types.go
```diff
@@ -3,219 +3,37 @@ package application
...
-	},
-	"ingress": {
-		"items": {
-			"ui:order": [
-				"enabled",
-				"host",
-				"path"
-			]
-		}
+func (app *Application) AsBundle() *bundle.Bundle {
```
coryodaniel, line 25: noice.

pkg/bundle/hydrate.go
```diff
@@ -41,6 +41,11 @@ func (b *Bundle) Hydrate(path string, c *client.MassdriverClient) error {
 		return err
 	}
 
+	if b.Connections == nil {
```
coryodaniel, line 44: We might need to do the same for Artifacts, id double check a yaml w/o an `artifacts` key.

For context, charts like Timescale and Clickhouse will have artifacts, but end user apps _probably_ wont until we release the web/api service artdef. We need to generalize there a bit, so I could easily see a fair amount of apps (including the massdriver platform-api app) having no `artifacts` set in the YAML file.

pkg/bundle/prompt.go
```diff
@@ -105,3 +112,52 @@ func getOutputDir(t *template.Data) error {
 	t.OutputDir = result
 	return nil
 }
+
+// TODO fetch these from the API instead of hardcoding
+func GetConnections(t *template.Data) error {
```
coryodaniel, line 117: If we dont have an artficat definitions list function now on the REST API, when we expose `deploy` function from the graph next week _all_ functions will be exposed, so we'll be able to make a graph call from here to get that list.

pkg/bundle/prompt.go
```diff
@@ -105,3 +112,52 @@ func getOutputDir(t *template.Data) error {
...
+	artifacts, err := common.GetMassdriverArtifacts()
```
coryodaniel, line 118: We should be careful to refer to them as artifact definitions or definitions since they aren't an artifact yet. Once the graph is exposed, i could see value in `mass artifact get ecomm-prod-cluster-0000 --field k8s_authentication --format kubeconfig | raw > ~/.kubeconfig` and it would be nice for the go code not to be overloading the word "artifact"

pkg/bundle/prompt.go
```diff
@@ -105,3 +112,52 @@ func getOutputDir(t *template.Data) error {
...
+	if err != nil {
+		return err
+	}
+	var selectedDeps []string
+	multiselect := &survey.MultiSelect{
+		Message: "What artifacts does your application depend on? If you have no dependencies just hit enter or only select (None)",
```
coryodaniel, line 124: If we are including this prompt on bundles now, `application` might be confusing here. Can we word the sentence without `app` or `bundle` or possible take in the `type` and interpolate?

pkg/bundle/prompt.go
```diff
@@ -105,3 +112,52 @@ func getOutputDir(t *template.Data) error {
...
+		}
+
+		validate := func(input string) error {
+			if !connectionNameFormat.MatchString(input) {
+				return errors.New("name must be at least 2 characters, start with a-z, use lowercase letters, numbers and underscores. It can not end with an underscore")
+			}
+			return nil
+		}
+
+		log.Info().Msgf("Please enter a name for the dependency %v", v)
```
coryodaniel, line 147: I never know the right level of information to present, we might be able to clarify this in the docs (and hopefully they've read those if they are making a bundle|app), but should we say something like the following or is it too wordy.

`Please enter a name for the dependency. This will be the variable name used to reference it in your app|bundle IaC`

pkg/bundle/prompt.go
```diff
@@ -105,3 +112,52 @@ func getOutputDir(t *template.Data) error {
...
+			Label:    `Name`,
+			Validate: validate,
+		}
+
+		result, errName := prompt.Run()
+		if errName != nil {
+			return errName
+		}
+
+		depMap[result] = fmt.Sprintf("massdriver/%s", selectedDeps[i])
```
coryodaniel, line 158: When we pull this from the api we'll need to remove `massdriver/` from it as it will have prefixes on it already, including private bundles like `evilcorp/iis-server-4`

pkg/bundle/templates/terraform/massdriver.yaml
```diff
@@ -26,16 +26,20 @@ params:
...
-  # properties:
-  #   aws_authentication:
-  #     $ref: massdriver/aws-iam-role
-  #   network:
-  #     $ref: massdriver/aws-regional-cloud
+<md end -md>
+
+<md if .Connections md>
+connections:
+  required: ["*"]
```
coryodaniel, line 36: '*' doesnt https://www.jsonschemavalidator.net/s/BnUUyUot, i used this as short hand the other day, sorry. I wish :D

We'll need to https://shopify.github.io/liquid/filters/join/ `.Connections`

pkg/common/artifacts.go
```diff
@@ -1,4 +1,4 @@
-package application
+package common
```
coryodaniel, line 1: leet

pkg/common/constants.go
```diff
@@ -43,3 +44,15 @@ const (
 	AllRW      = AllRead | AllWrite
 	AllRWX     = AllRW | AllExecute
 )
+
+var FileIgnores []string = []string{
```
coryodaniel, line 48: Definitely not something for now, but we should add a backlog item to make a `.massignore`. We'll probably want to exclude these forever in the cli, but devs may want to exclude other files in the directory.

coryodaniel (changes requested):
issa bundle

A few notes, consider approved after addressing.

pkg/bundle/templates/terraform/massdriver.yaml
WillBeebe: Lol I see... Fixed :)

pkg/common/constants.go
WillBeebe: Ticket created.

pkg/bundle/prompt.go
WillBeebe: Oh dopeee, that's going to be another game-changer.

pkg/bundle/prompt.go
WillBeebe: Good point, thank you

pkg/bundle/prompt.go
WillBeebe: Good point, I remember we talked about artifact / artifact-definition / etc.. Thanks always for the reminder, I'll update this to be accurate

pkg/application/types.go
WillBeebe: Good q. We've migrated some arrays to maps in TF to avoid the whole "recreates resources when the index in the array changes". However, we don't use policies _directly_. We use the policies to grab values from the `security` block and build a map of policy + iam. Since we for_each that map to create TF resources, I think we're good leaving Policies as an array in the CLI.

cmd/bundle.go
WillBeebe: Created a ticket to come back to this, will be a good addition.

WillBeebe:
> ![issa bundle]
> 
> A few notes, consider approved after addressing.

@coryodaniel feedback addressed

coryodaniel (approved):

PR #93, opened 2022-08-17
make template pattern more appealing

WillBeebe:
I thought `<md` in our templates looked a little gross, thoughts on `{{mass`? I added a test w/ a helm Chart using `{{` and it didn't have an issue.

### DIFF
--- a/pkg/bundle/templates/terraform/README.md
+++ b/pkg/bundle/templates/terraform/README.md
@@ -1,6 +1,6 @@
-# <md .Name md>
+# {{mass .Name mass}}
 
-<md .Description md>
+{{mass .Description mass}}
 
 ## Development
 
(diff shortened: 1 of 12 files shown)


coryodaniel:
I feel like a full word is confusing. I think we only need to avoid golang, helm, and any other IaC delims. We should build a list of those from the top few IaC tools and see if we can come up with a symbol that works for us

coryodaniel:
* {% - jinja2 / ansible - https://jinja.palletsprojects.com/en/3.0.x/templates.html
* ${ - terraform
* {{ - helm

I wonder if we could go the ruby route and do: `#{foo}` I’m not sure if the delims in the template enginee can handle non symmetric w a bunch of code with }}{}{{}} everywhere, so we may need to run a test real quick 

If taking a language delim is concerning maybe  %{} or $<>.

jaketf:
my .02, personally I actually like `<md md>` the best of all those suggested

coryodaniel:
`<md md>` is the least likely to intersect w/ another template engine.

coryodaniel:
Lets close this. `<md md>` can suffice until people tell us they no me gusta.

PR #95, opened 2022-08-17
remove publish path

WillBeebe:


coryodaniel (approved):
Fast track, demo in 18m!

PR #99, opened 2022-08-17
remove setting additional props on transforms

WillBeebe:
There were some comments in the method I removed

- // json schema has a bug where if "anyOf", "allOf" or "oneOf" are used, additionalProperties *MUST* be true
- https://github.com/massdriver-cloud/xo/issues/53

I thought additionalProperties could be false or an object. The other link points to this page, which doesn't talk about this at all, maybe it's been addressed in JSON Schema?

https://json-schema.org/understanding-json-schema/reference/combining.html#:~:text=biggest%20surprises


chrisghill (approved):
We can also remove the `md_set_id` transformation now that it's no longer used anywhere (and I think the front end team is ripping out the generation logic).

Doesn't have to be this PR if you want to keep them scoped separately.

PR #106, opened 2022-08-21
All / deny lists for publishing

WillBeebe:
For apps and bundles

- build -> builds in same directory
- publish -> builds, then copies to a tmp dir, tars the dir and sends to s3

During the copy, all files / dirs at the same level of the `massdriver.yaml` are put through an allowlist. We _only_ copy files or folders in this allowlist.

Once copying one of these allowed folders, we explicitly ignore some files and do some other checks on things like filesize.

Max single file size: 1MB
Max entire bundle size: 10MB


WillBeebe:
I locally ran this against GKE directing output to a `_build` directory so I could inspect it. 

```
ls -al core-services/
total 88
drwxr-xr-x  13 wbeebe  staff   416 Jul 14 11:39 .
drwxr-xr-x  18 wbeebe  staff   576 Aug 19 22:22 ..
drwxr-xr-x   4 wbeebe  staff   128 Jun 11 11:29 .terraform
-rw-r--r--   1 wbeebe  staff  5682 Jun 11 11:29 .terraform.lock.hcl
-rw-r--r--@  1 wbeebe  staff  1285 Jul  8 12:51 _artifacts.tf
-rw-r--r--@  1 wbeebe  staff   157 Aug 21 13:00 _connections_variables.tf.json
-rw-r--r--@  1 wbeebe  staff    89 Aug 21 13:00 _md_variables.tf.json
-rw-r--r--@  1 wbeebe  staff   359 Aug 21 13:00 _params_variables.tf.json
-rw-r--r--@  1 wbeebe  staff  1491 Jul  8 12:51 _providers.tf
-rw-r--r--   1 wbeebe  staff  3181 Jun 17 17:39 core_services.tf
-rw-r--r--   1 wbeebe  staff   400 Jun  9 21:25 locals.tf
-rw-r--r--   1 wbeebe  staff   369 Jun 17 17:39 observability.tf
-rw-r--r--   1 wbeebe  staff   943 Jun  9 21:25 service_account.tf

ls _build/
core-services           massdriver.yaml         schema-connections.json schema-ui.json
custom-resources        schema-artifacts.json   schema-params.json      src

ls _build/core-services/
_artifacts.tf      _providers.tf      core_services.tf   locals.tf          observability.tf   service_account.tf
```

pkg/common/constants.go
```diff
@@ -47,15 +49,25 @@ const (
...
-	".git",
+}
+
+var FileAllows []string = []string{
+	"massdriver.yaml",
+	"schema-params.json",
+	"schema-connections.json",
+	"schema-artifacts.json",
+	"schema-ui.json",
+	"src",
```
coryodaniel, line 72: Do we look for `steps` to see if there are any other directories to load?

WillBeebe: We sure do. If steps are not nil we add all steps to the allow list and dedupe

coryodaniel (changes requested):
Im not sure whats up, but this does publish (and `mass ___ build` looks right), but provisioning fails due to missing variables. Depending on the run connections are missing or md_metadata is missing. I ran the provisioner for other bundles that weren't published w/ this branch and they work.

Maybe we should make a local `_build` directory in the repo where `massdriver.yaml` is that isn't a tmp dir (is gitignored in our .gitignore template) so we can see whats in this directory.

coryodaniel (changes requested):
I can't add a suggestion because the files aren't editable, Im not sure how this worked before, but the following files _need_ to be in the bundle:

```
	// // Auto generated terraform variable files
	// "_connections_variables.tf.json",
	// "_params_variables.tf.json",
	// "_md_variables.tf.json",

**Comment by coryodaniel** [comment]
I commented those out locally and the build worked.

**Comment by WillBeebe** [comment]
> I commented those out locally and the ~build~ **PUBLISH** worked.

I lost my way a bit in the allow / ignore and order of operations but 🤦 duh we need those! I commented them out w/ a danger note for future devs, and added a test.

**Comment by coryodaniel** [comment]
I’d remove them. They should never be I commented.

Cory ODaniel
Mobile: 310.606.0257
***@***.***
medium: @coryodaniel ***@***.***>
twitter: @coryodaniel <https://twitter.com/coryodaniel>

File: pkg/common/constants.go (hunk 1 above)
Comment (coryodaniel) [reply pkg/common/constants.go:74]: 🚀

File: pkg/common/copy.go [hunk 2]
```diff
@@ -0,0 +1,121 @@
...
+		if errCopy != nil {
+			return stats, errCopy
+		}
+	}
+
+	return stats, nil
+}
+
+// Written because path.Walk traverses the entire directory tree.
+// If a step has a folder like .terraform, path.Walk will still traverse it
```
jaketf, line 49: hmm, I think it will be less code / more maintainable if we use the canonical Walk pattern. 
You can avoid walking a dir by your `WalkFunc` returning the special error `filepath.SkipDir`. 

from path/filepath godocs [here](https://pkg.go.dev/path/filepath#pkg-variables:~:text=SkipDir%20is%20used%20as%20a%20return%20value%20from%20WalkFuncs%20to%20indicate%20that%20the%20directory%20named%20in%20the%20call%20is%20to%20be%20skipped.%20It%20is%20not%20returned%20as%20an%20error%20by%20any%20function.):

> SkipDir is used as a return value from `WalkFunc`s to indicate that the directory named in the call is to be skipped. It is not returned as an error by any function.

you can see how it is used in the [filepath.Walk example](https://pkg.go.dev/path/filepath#example-Walk)

pkg/bundle/testdata/zipdir/schema-ui.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$id": "https://schemas.massdriver.cloud/schemas/bundles/gcp-gke-cluster/schema-ui.json",
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "properties": {},
+    "title": "gcp-gke-cluster"
+}
```
chrisghill, line 6: Leaving the comment here, cuz I can't comment on the empty file.

I just want to make sure the file below (`.terraform/stuff.tf`) is there to test **ignore** functionality. We def don't want `.terraform` in the bundle package.

pkg/common/constants.go
```diff
@@ -47,15 +49,27 @@ const (
...
-	"_md_variables.tf.json",
-	"_params_variables.tf.json",
+	// Auto generated terraform variable files
+	// We need to include these. Comment left
+	// so we don't try to ignore them in the future
+	// "_connections_variables.tf.json",
+	// "_params_variables.tf.json",
+	// "_md_variables.tf.json",
+	// For setting variable input values during _local_ development
+	"_connections.auto.tfvars.json",
```
chrisghill, line 57: We can probably just do `tfvars.json`.  If somebody accidentally misspells `_conections.auto.tfvars.json` we don't want to let that through.

Edit: should do `.tfvars` and `.tfvars.json` in case some users set variables without using json.

jaketf (changes requested):

WillBeebe:
Ticket [here](https://github.com/massdriver-cloud/massdriver/issues/1297) and I updated the file in the PR. 💪

pkg/common/constants.go
```diff
@@ -47,15 +49,27 @@ const (
...
-	".git",
+}
+
+var FileAllows []string = []string{
+	MassdriverYamlFilename,
+	ArtifactsSchemaFilename,
+	ConnectionsSchemaFilename,
+	ParamsSchemaFilename,
+	UISchemaFilename,
+	"src",
```
chrisghill, line 68: Are we having folders in the `FileAllows` list? This becomes unclear exactly what this list is for.

At first I assumed it was looking for filenames that matched the string. What does it do w/ directories?

pkg/common/copy.go
WillBeebe: That would have saved some time. :) Do you think this is blocking or can we come back to it?

pkg/common/copy.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// a file we should include, write file
+	if !info.IsDir() {
+		stats.FolderSize += info.Size()
+		data, err1 := ioutil.ReadFile(src)
+		if err1 != nil {
+			return err1
+		}
+
+		return ioutil.WriteFile(dest, data, AllRWX)
```
chrisghill, line 69: It would be great to retain `chmod` settings on the copy over if that's possible, instead of making everything `777`. If we can't, honestly we might want to just do `UserRW`. The provisioning runs as the `massdriver` user, and when it unzips it should inherit the "user" permissions. Nothing else should need to read, write, or execute the file.

pkg/common/copy.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	// a folder we should include
+	// so we create the folder, then iterate through
+	// the files in that folder.
+	errMkdir := os.Mkdir(dest, AllRX|UserRW)
```
chrisghill, line 75: Same with directories, but I'm less concerned about dir permissions.

pkg/common/copy.go
```diff
@@ -0,0 +1,121 @@
...
+
+	for _, ignore := range config.Ignores {
+		if strings.Contains(fileName, ignore) {
+			return true
+		}
+	}
+
+	mbs := FileSizeMB(info.Size())
+	if mbs > MaxFileSizeMB {
+		log.Debug().Msgf("File: %s is larger than limit of %vMB.", fileName, MaxFileSizeMB)
```
chrisghill, line 116: Should make this an `Error` or at least a `Warning`. This will effectively be a *silent* failure if its hidden in debug logs.

pkg/common/copy_test.go
```diff
@@ -0,0 +1,72 @@
...
+			if err != nil {
+				t.Fatalf("%d, unexpected error", err)
+			}
+
+			gotMD5, err := dirhash.HashDir(path.Join(testDir), "", dirhash.DefaultHash)
+			if err != nil {
+				t.Fatalf("%d, unexpected error", err)
+			}
+
+			if gotMD5 != wantMD5 {
```
chrisghill, line 67: I wish there was a better way to check folder equality than these fragile MD5 hashes, bit it works for now.

pkg/common/copy.go
WillBeebe: Alright, I'll make these changes

chrisghill (commented):
Looks good. Some minor feedback and naming suggestions.

pkg/bundle/testdata/zipdir/schema-ui.json
WillBeebe: Yes I explicitly named it `.terraform` to make sure it's excluding in the Package phase.

pkg/common/constants.go
coryodaniel: ^ yes.

pkg/common/copy_test.go
coryodaniel: There was a walk and diff I added for app templates... https://github.com/massdriver-cloud/massdriver-cli/blob/971d2723caefcff297b2a6aeb673852acd4dc217/pkg/template/render_test.go#L83

pkg/common/constants.go
WillBeebe: Okay, any filename that contains `.tfvars` is ignored by one of the earlier file ignores. I yanked these explicit ones.

pkg/common/constants.go
WillBeebe: It follows the same logic. If the string we want to copy (file or folder) matches any of these, we include it.

pkg/common/copy.go
WillBeebe: Great point, thank you

pkg/bundle/package.go
```diff
@@ -3,13 +3,56 @@ package bundle
...
+	if errCopy != nil {
+		return errCopy
+	}
+
+	mbs := common.FileSizeMB(stats.FolderSize)
+	if mbs > common.MaxBundleSizeMB {
+		return fmt.Errorf("Bundle size exceeds maximum allowed size of %vMB", common.MaxBundleSizeMB)
+	}
+
+	return tarFolder(buildDir+"/"+common.MassdriverYamlFilename, buf)
```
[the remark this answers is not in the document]
WillBeebe: Good point, this was legacy logic as things were being consolidated. I can drop the filename now and just pass the dir.

pkg/bundle/package.go
```diff
@@ -3,18 +3,60 @@ package bundle
...
+	"os"
 	"path/filepath"
 
+	"github.com/massdriver-cloud/massdriver-cli/pkg/common"
 	"github.com/massdriver-cloud/massdriver-cli/pkg/compress"
 )
 
-func PackageBundle(filePath string, buf io.Writer) error {
+func Package(b *Bundle, filePath string, buf io.Writer) error {
+	buildDir, err := os.MkdirTemp("", "bundle-build")
```
jaketf, line 16: should this temp dir include the bundle name (or at least a * to add some randomness) to avoid temp dir name collision? kind of a weird scenario but if you were packaging multiple bundles with different names concurrently (e.g. in CI) it'd be a frustrating if this temp dir name made that not possible.

pkg/bundle/package.go
```diff
@@ -3,18 +3,60 @@ package bundle
...
+	mbs := common.FileSizeMB(stats.FolderSize)
+	if mbs > common.MaxBundleSizeMB {
+		return fmt.Errorf("Bundle size exceeds maximum allowed size of %vMB", common.MaxBundleSizeMB)
+	}
+
+	return tarFolder(buildDir, buf)
+}
+
+func getAllowList(b *Bundle) []string {
+	allAllows := []string{}
```
jaketf, line 42: non-blocking educational nit: could be more efficient we know the  length of this slice apriori so could pre-allocate.
though this might be slightly less readable.
```suggestion
	allAllows := make([]string, len(common.FileAllows) + len(b.Steps))
	copy(allAllows, common.FileAllows)
	if b.Steps != nil {
	  for i, step := range b.Steps {
	    allAllows[len(common.FileAllows)+i] = step.Path
	  }
	}
```

pkg/bundle/package_test.go
```diff
@@ -16,25 +16,43 @@ func TestPackageBundle(t *testing.T) {
...
 			var got bytes.Buffer
-			err := bundle.PackageBundle(tc.bundlePath, &got)
+			err := bundle.Package(tc.bundle, tc.bundlePath, &got)
 			if err != nil {
 				t.Fatalf("%d, unexpected error", err)
 			}
 
 			// Create a temp dir, write out the archive, then shell out to untar
 			testDir := t.TempDir()
+			// testDir := "_build"
```
jaketf, line 55: remove commented out code which I assume is a debugging artifact

pkg/common/utils.go
```diff
@@ -62,3 +30,25 @@ func WriteFile(filePath string, data []byte, errToBytes error) error {
...
 	return nil
 }
+
+// TODO: use generics
+func RemoveDuplicateValues(stringSlice []string) []string {
+	keys := make(map[string]bool)
+	list := []string{}
+
+	for _, entry := range stringSlice {
+		if _, value := keys[entry]; !value {
```
jaketf, line 40: this "value" name is low-key really misleading because it's actually a bool that indicates if the key was found _not_ the value for that key... (though the way you're using this data structure those will _happen_ to be the same value if we ever introduced a bug in this code it'd be a real head scratcher)
In most go code I've seen the common idiom when using the dual assignment from a map key lookup is to use the name `ok` in this scenario.
```suggestion
		if _, ok := keys[entry]; !ok {
```

that being said I don't event think that is necessary here as map[string]bool will return the zero value  false if the key is not found so the following would be a more succinct way to express what's going on here. 
```suggestion
		if keys[entry] {
```
finally I think keys should be called something like `seen` to indicate these are the unique values we've already seen.

pkg/common/utils.go
```diff
@@ -62,3 +30,25 @@ func WriteFile(filePath string, data []byte, errToBytes error) error {
...
+			keys[entry] = true
+			list = append(list, entry)
+		}
+	}
+
+	return list
+}
+
+const tenTwentyFour = 1024
```
jaketf, line 49: lol was this to make the linter happy?

I typically see these magic numbers named something more descriptive.
Also there's no reason for the extra memory allocation / math operation of calcuating kilobytes if you only care about megabytes.
```suggestion
const bytesInMB = 1024 * 1024
```

jaketf (approved):
some small suggestions otherwise looks good! thanks for making the WalkFunc change!

coryodaniel:
PS, ive packaged w/ this branch like 10 times today and have had no issues.

pkg/common/utils.go
WillBeebe: I didn't know I could do that for a function that returns two values! V cool. I rewrote and renamed this locally 👍

pkg/common/utils.go
WillBeebe: Lol yes. Thanks for the code suggestion.

pkg/bundle/package.go
WillBeebe: Since the number of elements are so low I think I'll favor the readability here but you make a great point.

pkg/bundle/package.go
WillBeebe: I'm pretty sure it already does add a random suffix.

https://pkg.go.dev/os#MkdirTemp

If we wanted it at a different position in the string we could do `bundle-*-build` but I'm pretty sure as-is it will be `bundle-build128381910`. LMK what you think.

WillBeebe: Also, this might just be a readability thing. I'll use `bundle-build-*` so it's more clear. 👍

coryodaniel (approved):

PR #107, opened 2022-08-22
Support for overriding App template reader func for local tmpl dev

coryodaniel:
* [Changed] cache to use private functions to access the cache for manipulating files.
* [Added] `MD_DEV_TEMPLATES_PATH` that will alter the templates reader for local development of templates

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -26,8 +26,9 @@ var applicationCmd = &cobra.Command{
 }
 
 var applicationBuildCmd = &cobra.Command{
-	Use:  "build",
-	RunE: runApplicationBuild,
+	Use:   "build",
+	Short: "Builds the app for local development",
+	RunE:  runApplicationBuild,
 }
 
 var applicationNewCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


cmd/application.go
```diff
@@ -161,3 +168,18 @@ func runTemplatesRefresh(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+func runTemplatesList(cmd *cobra.Command, args []string) error {
```
WillBeebe, line 172: This does the same thing w/ different logging as this method.

https://github.com/massdriver-cloud/massdriver-cli/blob/main/cmd/application.go#L142-L152

```
mass app templates 
mass app templates list
```

We should combine the listing logic or change what `mass app templates` does, maybe just usage if it has a subcommand?

pkg/cache/application.go
```diff
@@ -12,9 +12,20 @@ import (
 // TODO: pull from massrc
 const MassdriverApplicationTemplatesRepository = "https://github.com/massdriver-cloud/application-templates"
 
+// AppTemplateCacheDir is a reader function to access the local cache of templates.
+// When developing templates, the cache source can be overwritten for reads by setting `MD_DEV_TEMPLATES_PATH`
 func AppTemplateCacheDir() string {
```
WillBeebe, line 17: nice and tidy 👍

WillBeebe (approved):
Only question one question otherwise LGTM

cmd/application.go
coryodaniel: I didn’t see this cmd!

should I yank this and move the other command here?

WillBeebe: What about removing the other commands function and the call to it on line 50? So that "mass app templates" wouldn't list anymore but it would show the available subcommands.

coryodaniel:
I"m going to skip that one failing lint rule ~FileIgnores so you don't end up with a conflict.

PR #108, opened 2022-08-24
Bugfix/copy local

WillBeebe:
Issue:

When I moved from my recursive copy method to using filepath.Walk I introduced a bug. `filepath.Walk` behaves differently when copying from local root vs non-local root. 

```
// tests
copyFrom: testdata/bundle
relative path: /

// real-world bundle build 
copyFrom: .
relative path: .
```

Because this behaves differently for local, the tests didn't catch it. 

This wasn't fun: 
Write code -> test with bundle -> write test -> rewrite code -> tests pass -> declare victory -> accept defeat. 

So I tried this:
Write test that captures the bug -> fix code -> test with bundle 

Here's a run with the new test and without the changes (test fails).

```
git stash
Saved working directory and index state WIP on copy-local: 66eafa9 add test for local

make test
go test ./cmd
?   	github.com/massdriver-cloud/massdriver-cli/cmd	[no test files]
go test ./pkg/...
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/application	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/bundle	0.221s
?   	github.com/massdriver-cloud/massdriver-cli/pkg/cache	[no test files]
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/client	(cached)
Comparing (want) testdata/copy-from-root/utils.go and (got) utils.go
=865
--- FAIL: TestCopyFolder (0.00s)
    --- FAIL: TestCopyFolder/CopyFromRoot (0.00s)
        copy_test.go:86: got h1:47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=, want h1:uvsIsiJ8JZ/oX+4EsTYQJHsLIZsW2r4NqSiT0pl1mjA=
FAIL
FAIL	github.com/massdriver-cloud/massdriver-cli/pkg/common	0.122s
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/compress	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/definition	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/jsonschema	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/provisioners/terraform	0.135s
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/template	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/version	(cached)
FAIL
make: *** [test] Error 1
```

New test, with the changes

```
git stash pop

make test
go test ./cmd
?   	github.com/massdriver-cloud/massdriver-cli/cmd	[no test files]
go test ./pkg/...
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/application	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/bundle	0.420s
?   	github.com/massdriver-cloud/massdriver-cli/pkg/cache	[no test files]
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/client	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/common	0.338s
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/compress	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/definition	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/jsonschema	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/provisioners/terraform	0.350s
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/template	(cached)
ok  	github.com/massdriver-cloud/massdriver-cli/pkg/version	(cached)
```


pkg/common/copy_test.go
```diff
@@ -50,6 +50,17 @@ func TestCopyFolder(t *testing.T) {
 				Ignores: []string{},
 			},
 		},
+		{
+			name:       "CopyFromRoot",
+			bundlePath: ".",
```
WillBeebe, line 55: this is a really lame and fragile way to test this! I'm not sure how to mock / stub this the right way but this should definitely be replaced ASAP.

pkg/common/testdata/copy-from-root/utils.go
```diff
@@ -0,0 +1,53 @@
+package common
```
WillBeebe, line 1: the expected output from the lame test above

coryodaniel:
What would be nice _on publish_ would be to use the allowlist of files (yaml, schemas, step dirs) and validate that those _are_ in the bundle and fail if they are not.

I wouldn't add it to this branch unless that's a super quick change as I'd like to get this merged pronto. FF to add a ticket for the above if its not a fast one.

pkg/common/copy.go
```diff
@@ -28,9 +28,16 @@ func CopyFolder(src, dst string, config *CopyConfig) (CopyStats, error) {
 		if err != nil {
 			return err
 		}
+		if path == "." {
```
jaketf, line 31: I'm not sure I follow why this case returns nil. could we explain in a comment?

pkg/common/copy.go
```diff
@@ -28,9 +28,16 @@ func CopyFolder(src, dst string, config *CopyConfig) (CopyStats, error) {
...
+			return nil
+		}
+
 		relPath := strings.Replace(path, src, "", 1)
-		depth := strings.Count(relPath, string(os.PathSeparator))
+		if src == "." {
```
jaketf, line 36: I feel like you're reaching for / re-implementing  [`filepath.Rel`](https://pkg.go.dev/path/filepath#Rel) here. Is there a reason this standard library method doesn't fit your needs?

jaketf: should be something like
```go
relPath, err = filepath.Rel(src, path)
```

pkg/common/copy.go
```diff
@@ -39,8 +46,9 @@ func CopyFolder(src, dst string, config *CopyConfig) (CopyStats, error) {
 			return nil
 		}
 
+		writePath := dstDir + relPath
```
jaketf, line 49: it's always better form when working with paths to use the path manipulation methods from standard library to avoid issues with leading / trailing slashes.
```suggestion
		writePath := filepath.Abs(filepath.Join(dstDir, relPath))
```

pkg/common/copy.go
jaketf: also, in general it might make our life more sane to debug if we just immediately do a `absPath := filepath.Abs(path)` and then deal with the abs path from there.

pkg/common/copy_test.go
jaketf: why is this called "copy from root" that makes me think it's copy from `/` but I think what you're testing is "copy from working directory" is that correct?

jaketf (changes requested):

pkg/common/copy_test.go
jaketf: I think the "best way to do this" to do something like copying from `.` is:
```go
// TODO handle these errs
originalDir, err := os.Getwd()
err := os.Chdir("some/test/bundle/path")
// change back to original dir after this test function completes
defer func(dir string) {
  if err := os.ChDir(dir); if err != nil {
    panic(err)
  }
}(originalDir)
// do the rest of your test you're now in the bundle dir and can refer to it as "."
```

pkg/common/copy.go
WillBeebe: Simply because I had no idea that existed. Thanks for the help 👏

WillBeebe:
@jaketf and @coryodaniel this is ready to go. The test is now sane and I ran through the CLI locally publishing a bundle to my private org.

coryodaniel (approved):
Packaged & provisioned.

pkg/common/copy_test.go
```diff
@@ -90,6 +79,52 @@ func TestCopyFolder(t *testing.T) {
...
+		},
+		Ignores: []string{},
+	}
+
+	_, err2 := common.CopyFolder(".", copyTo, config)
+	if err2 != nil {
+		t.Fatalf("%d, unexpected error", err2)
+	}
+	// change back to the original dir so want paths and got paths make sense
+	if err := os.Chdir(originalDir); err != nil {
```
jaketf, line 108: FWIW if you hit any of the `t.Fatalf`s before this I think you won't go back to this original Dir that that's one advantage of putting a `defer` in as a failsafe so your test implementation of changing directories is not leaked into confusing issues.

jaketf (approved):
LGTM! made one note about dir changes in the test that doesn't have to block merging this PR if the test seems to be working we can clean it up later if necessary.

Issue #114, opened 2022-08-29
Make the commands make more sense.


coryodaniel:
Now that we are starting to expose some of the other graph functions, particularly `package` functions like `deploy`... our commands are a bit confusing.

* `mass app new` makes a application bundle (source code)
* `mass app publish` publishes the bundle to our _package_ manager
* `mass deploy` -> doesn't feel right as a top-level command since it technically 'deploys a package'...
* `mass package deploy` starts to sound really confusing wrt to package management (these are deployed resources), as would `mass package configure`

Since applications and bundles don't really have any differences in how they are execute, I propose we move them into just `bundles` subcommand. The only difference in how we managed the bundles is how they are templatized.

* mass bundle new - will make a new bundle from a template
  * our terraform template will need to move to `application-templates` (we should also rename this to `bundle-templates`)
  * terraform template will need to be updated to support `liquid`
* mass bundle publish - will publish a bundle (mind you "bundle" is becoming our generic term for a bag of IaC, current "bundle" types are going to be changed to "infra" or "infrastructure"

This will free up `mass app` to deal w/ end user applications:

`mass app deploy ecomm-staging-api` will deploy the application package
`mass app config ecomm-staging-api -f ./params.json` would read JSON from file and configure the package for deployment

WillBeebe:
"terraform template will need to be updated to support `liquid`" => already done, I grabbed this as part of the app templating

WillBeebe:
I like `mass app deploy` and `mass app config`, and it makes sense to consolidate on `bundle` now. 👍

PR #115, opened 2022-08-29
mass deploy

coryodaniel:
Temporarily making this a top level command until addressed #114

* [Added] `mass deploy _naming_convention_` i.e.: `mass deploy ecomm-prod-web` (note: not suffix)
* [Added] `mass --json` to enable JSON output
* [Changed] `mass --verbose` to set INFO level if not verbose


pkg/common/constants.go
```diff
@@ -47,7 +47,7 @@ const (
 	AllRWX     = AllRW | AllExecute
 )
 
-var FileIgnores []string = []string{
+var FileIgnores = []string{
```
coryodaniel, line 50: Forced to make these changes, pre-commit kept failing.

pkg/api/package.go
```diff
@@ -0,0 +1,136 @@
...
+				ID graphql.String
+			}
+			ActiveDeployment struct {
+				ID     graphql.String
+				Status graphql.String
+			}
+			Target struct {
+				ID graphql.String
+			}
+		} `graphql:"getPackageByNamingConvention(name: $name, organizationId: $organizationId)"`
```
chrisghill, line 82: I've not seen this syntax before.  A string literal right after a struct declaration? How does this work?

chrisghill (approved):

pkg/api/http.go
```diff
@@ -0,0 +1,32 @@
+package api
```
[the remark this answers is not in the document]
coryodaniel: We'll hopefully be getting rid of the REST API at some point in the near future, and then we can. It looks like that library sets the header a custom header instead of an `authorization` header and it does it _as_ a request instead of _per request_. We don't have access the the HTTP call in the gql library, so in the impl here you'll see its set up as a roundtripper that modifies the request.

pkg/api/package.go
coryodaniel: you have seen it ... a struct is just a type like a string or int.

```go
type Foo struct {
  Name string `json:"name"`
}
```

chrisghill: ohhhhhh, I didn't know graphql had its on field tags. Nice!

PR #116, opened 2022-09-08
Hide alpha templates from display

coryodaniel:
Hides [`alpha`](https://github.com/massdriver-cloud/application-templates/tree/main/alpha/aws-ec2) templates from display.

To opt-in:

```shell
mass app template refresh
MD_DEV_TEMPLATES_PATH="~/work/application-templates/alpha" mass app new
```

### DIFF
--- a/pkg/application/prompt.go
+++ b/pkg/application/prompt.go
@@ -145,6 +145,7 @@ func getOutputDir(t *template.Data) error {
 	return nil
 }
 
+// note, does not preserve ordering
 func remove(s []string, i int) []string {
 	s[i] = s[len(s)-1]
 	return s[:len(s)-1]
--- a/pkg/application/prompt.go
+++ b/pkg/application/prompt.go
@@ -95,8 +95,12 @@ func getDescription(t *template.Data) error {
 	return nil
 }
 
+var IgnoredTemplateDirs = []string{"alpha"}
+
 func getTemplate(t *template.Data) error {
 	templates, err := cache.ApplicationTemplates()
+	templates = removeIgnoredTemplateDirectories(templates, IgnoredTemplateDirs)
+
 	if err != nil {
 		return err
 	}
@@ -113,6 +117,21 @@ func getTemplate(t *template.Data) error {
 
 	t.TemplateName = result
 	return nil
+}
+
+func removeIgnoredTemplateDirectories(templates []string, remove []string) []string {
+	for i := 0; i < len(templates); i++ {
+		template := templates[i]
+		for _, rem := range remove {
+			if template == rem {
+				templates = append(templates[:i], templates[i+1:]...)
+				i--
+				break
+			}
+		}
+	}
+
+	return templates
 }
 
 func getOutputDir(t *template.Data) error {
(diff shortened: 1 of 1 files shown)


pkg/application/prompt.go
```diff
@@ -115,6 +119,21 @@ func getTemplate(t *template.Data) error {
 	return nil
 }
 
+func removeIgnoredTemplateDirectories(templates []string, remove []string) []string {
```
jaketf, line 122: this behavior should have an accompanying unit test.

pkg/application/prompt.go
```diff
@@ -95,8 +95,12 @@ func getDescription(t *template.Data) error {
 	return nil
 }
 
+var IgnoredTemplateDirs = []string{"alpha"}
```
jaketf, line 98: the choice of list of string here makes the code that uses this value less performant / readable (see comment below)
```suggestion
var IgnoredTemplateDirs = map[string]bool{"alpha": true}
```

pkg/application/prompt.go
```diff
@@ -115,6 +119,21 @@ func getTemplate(t *template.Data) error {
 	return nil
 }
 
+func removeIgnoredTemplateDirectories(templates []string, remove []string) []string {
+	for i := 0; i < len(templates); i++ {
+		template := templates[i]
+		for _, rem := range remove {
```
jaketf, line 125: performance nit 1: this is inefficient you loop over every entry of remove (albeit for now only one entry) on every iteration.
would be faster to define a remove map (suggested above) and if each template is in that map

performance nit 2: this method of removing an element from an array is also inefficient as order is not important here. SO https://stackoverflow.com/questions/37334119/how-to-delete-an-element-from-a-slice-in-golang

readability nit: this would also remove the nested for loop which is more of a cognitive burden on the reader.

then this becomes
```go
// in this case order does not matter so 
// fastest thing to do is replace removed item with last 
// and drop the last
func remove(s []string, i int) []string {
  s[i] = s[len(s)-1]
  return s[:len(s)-1]
}

for i, v  := range templates {
  if IgnoreTemplateDirs[v] {
    templates = remove(templates, i)
  }
}
```
example in go playground https://go.dev/play/p/odd4Y1HDUEz

jaketf (changes requested):

pkg/application/prompt.go
coryodaniel: Going to punt on a test here. This is just so some stuff that isn't ready for public isn't showing up when people go through the tutorial. We have no tests around any of the app template stuff and we need to get the docs out ASAP.

I assume it'll all change greatly when we address https://github.com/massdriver-cloud/massdriver-cli/issues/114

pkg/application/prompt.go
```diff
@@ -149,3 +144,8 @@ func getOutputDir(t *template.Data) error {
 	t.OutputDir = result
 	return nil
 }
+
+func remove(s []string, i int) []string {
```
jaketf, line 148:
```suggestion
// note, does not preserve ordering
func remove(s []string, i int) []string {
```

jaketf (approved):

WillBeebe (approved):

PR #118, opened 2022-09-08
Replace `ref` with `source_url`

coryodaniel:
Replace `ref` with `source_url`. This is a breaking change, but ... lets get it out so the docs are right :D

### DIFF
--- a/pkg/template/render_test.go
+++ b/pkg/template/render_test.go
@@ -97,14 +97,15 @@ func walkAndCompare(wantDir string, gotDir string) {
 				return err
 			}
 
-			fmt.Printf("Comparing (want) %s and (got) %s\n", path, gotFilePath)
-
-			dmp := diffmatchpatch.New()
 			gotText, _ := readFile(gotFilePath)
 			wantText, _ := readFile(path)
-			diffs := dmp.DiffMain(wantText, gotText, false)
-
-			fmt.Println(dmp.DiffToDelta(diffs))
+			if gotText != wantText {
+				fmt.Printf("File did not render correctly: %s\n", path)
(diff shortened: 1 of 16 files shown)


pkg/template/render_test.go
```diff
@@ -97,14 +97,15 @@ func walkAndCompare(wantDir string, gotDir string) {
...
 
-			fmt.Printf("Comparing (want) %s and (got) %s\n", path, gotFilePath)
-
-			dmp := diffmatchpatch.New()
 			gotText, _ := readFile(gotFilePath)
 			wantText, _ := readFile(path)
-			diffs := dmp.DiffMain(wantText, gotText, false)
-
-			fmt.Println(dmp.DiffToDelta(diffs))
+			if gotText != wantText {
```
WillBeebe, line 102: 👍

pkg/template/testdata/application-templates/renders-successfully/massdriver.yaml
```diff
@@ -11,20 +11,9 @@ params:
       type: string
     app_env:
       type: string
-<md if not .Connections md>
```
WillBeebe, line 14: Interesting that the test passed with this removed. The test-case has a `draft_node` and the expected file does as well.

https://github.com/massdriver-cloud/massdriver-cli/pull/118/files#diff-26cd67a32b447e457298d13f30e1a4206249ff485a89b83029954cb7cdb16a65R15-R20

WillBeebe (approved):
LGTM

pkg/template/testdata/application-templates/renders-successfully/massdriver.yaml
coryodaniel: The test never added connections so this just ended up being a red herring of whitespace

PR #119, opened 2022-09-09
add nil protection in deployment error reporting

jaketf:
currently if there's no deployment to get from the backend for some reason the cli panics / crashes in deref of `deployment.ID`.

unfortunately, even when we add this nil protection this doesn't give a ton more useful error info..

```
./bin/mass-linux-amd64 deploy jakegcp-gcpplay-db-dj48
FTL Deployment failed error="Message: 500 Internal Server Error; body: \"{\\\"errors\\\":{\\\"detail\\\":\\\"Internal Server Error\\\"}}\", Locations: []"
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -47,7 +47,6 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-
 	log.Info().Str("deploymentId", deployment.ID).Msgf("Deployment %s", strings.ToLower(deployment.Status))
 	return nil
 }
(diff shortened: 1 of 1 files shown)


cmd/deploy.go
```diff
@@ -39,7 +39,11 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 	deployment, err := api.DeployPackage(client, orgID, name)
 
 	if err != nil {
-		log.Fatal().Err(err).Str("deploymentId", deployment.ID).Msg("Deployment failed")
+		if deployment != nil {
```
coryodaniel, line 42: Have you see both of these cases?

PR #135, opened 2022-09-29
fix: test for the case of missing params file

WillBeebe:
Small fix for if the params file isn't present, this fails. It creates an empty file, then tries to unmarshal a blank string into JSON. To reproduce `mass app new ; mass app build`

Here's how it fails in our CI pipelines: https://github.com/massdriver-cloud/gcp-memorystore-redis/actions/runs/3148677430/jobs/5119469261


pkg/provisioners/terraform/compile.go
```diff
@@ -142,14 +143,21 @@ func getExistingParams(path string) (map[string]interface{}, error) {
...
+
+	stat, statErr := os.Stat(abs)
+	if os.IsNotExist(statErr) {
 		// no existing params return empty map
 		return params, nil
 	} else if statErr != nil {
 		return params, statErr
 	}
+	if stat.Size() == 0 {
+		log.Info().Msg("size is 0")
```
WillBeebe, line 155:
```suggestion
	
```

pkg/provisioners/terraform/compile.go
```diff
@@ -142,14 +143,21 @@ func getExistingParams(path string) (map[string]interface{}, error) {
...
+	if os.IsNotExist(statErr) {
 		// no existing params return empty map
 		return params, nil
 	} else if statErr != nil {
 		return params, statErr
 	}
+	if stat.Size() == 0 {
+	
+		return params, nil
+	}
```
WillBeebe, line 157:
```suggestion
	// if the file exists but is empty
	if stat.Size() == 0 {
		return params, nil
	}
```

jaketf (approved):
good catch, thanks for the quick PR!

PR #138, opened 2022-10-12
fix so template builds on first run

WillBeebe:
`mass bundle new` then `mass bundle build` breaks, because the template doesn't have any named examples.


coryodaniel:
Nice. Do you even need the foo fields or just a __name?

coryodaniel (approved):

WillBeebe:
> Nice. Do you even need the foo fields or just a __name?

Just a `__name`. Admittedly we could have a better default than foo, maybe `region`, `disk_size_mb` for the infra bundle template.

PR #140, opened 2022-10-31
add more to default md_metadata

WillBeebe:
Some bundles expect `md_metadta.deployment.id`. Adding this to the boilerplate for our bundle devs.


akasprzok (approved):
Super quick turnaround

PR #141, opened 2022-11-01
add template file for alarms

WillBeebe:


WillBeebe:
Hella old, we’ll refresh templates

PR #147, opened 2022-11-29
using same go type for infrastructure and applications, misc 🧹 

WillBeebe:
- added methods `IsInfrastructure()` and `IsApplication` to the bundle
- removed pure functions that did this checking in the `cmd` package
- replaced those functions with calls to the new methods
- removed the `Application` type
- updated references to the shared type

It's backwards-compat and prints a warning if the bundle type is `bundle`.


WillBeebe:
blocked by: https://github.com/massdriver-cloud/massdriver-cli/pull/148

cmd/application.go
```diff
@@ -163,8 +158,8 @@ func RunApplicationDeploy(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	if errType := checkIsApplication(app); errType != nil {
-		return errType
+	if !app.IsApplication() {
+		return fmt.Errorf("this command can only be used with bundle type 'application'")
```
mclacore, line 162: Would it be better practice to set this error message as a `const` above and ref the `const` in each return? like
```
const onlyAppErr = "this command can only be used with bundle type 'application'"

if !app.IsApplication() {
    return fmt.Errorf(onlyAppErr)
```

cmd/bundle.go
```diff
@@ -89,8 +82,8 @@ func runBundleBuild(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	if errType := checkIsBundle(b); errType != nil {
-		return errType
+	if !b.IsInfrastructure() {
+		return fmt.Errorf("this command can only be used with bundle type 'infrastructure'")
```
mclacore, line 86: Same thing here, but for infra

pkg/application/parse_test.go
```diff
@@ -5,19 +5,20 @@ import (
...
 	}
 	tests := []test{
 		{
 			name:    "app-spec",
 			appPath: "./testdata/massdriver.yaml",
-			want: application.Application{
+			want: bundle.Bundle{
 				Schema:      "draft-07",
 				Name:        "my-app",
 				Description: "An application",
```
mclacore, line 24: NIT: I know this is just a test, but would we want to leave the `name`, `Name`, and `Description` related to apps rather than bundles?

pkg/bundle/parse.go
```diff
@@ -15,29 +16,33 @@ type Overrides struct {
 // overrides allow the CLI to override specific bundle metadata.
 // This is useful in a CI/CD scenario when you want to change the `access` if you are deploying to a sandbox org.
 func Parse(path string, overrides map[string]interface{}) (*Bundle, error) {
-	bundle := new(Bundle)
+	b := new(Bundle)
```
mclacore, line 19: Just for my own learning, what is the purpose for replacing `bundle` with `b`?

mclacore (commented):
I'm surprised I actually understood ~50% of this PR lol. Just some questions about how/why for certain things, mostly for my learning.

pkg/bundle/parse.go
coryodaniel: save keystrokes, not a fan personally.

coryodaniel (approved):
Let’s go ahead and merge, not sure bikeshedding on IsApp is warranted given we are going to be moving all bundle management under “bundle” re: #114

cmd/bundle.go
WillBeebe: That's fair. I think the golang pattern here is to create an `errors.go` where the `fmt.Error(....` lives, so this just returns a constant.

pkg/application/parse_test.go
WillBeebe: So with this change, the underlying golang struct is a `Bundle`. It has a property named `Type` that can be infrastructure, bundle (legacy), or application.

pkg/bundle/parse.go
WillBeebe: Yeah, I hated this when I first started writing golang. These changes are to make this file consistent. 🤷 With `mass` we'll have consistent patterns 👍 .

PR #149, opened 2022-12-24
Adding chained arguments to CLI

likelymichael:
Still a WIP. Trying to get `mass bundle new` arguments working first, then I'll move onto other commands. The connections and connection names are by far the trickiest part.


mclacore:
Also took the liberty of removing the old `mass bundle generate` command since it's depreciated. This decision came mostly because the `-output-dir` argument was conflicting with the argument I was trying to make for `mass bundle new`.

PR #150, opened 2023-01-05
adding app block to the publish call. changing app block to pointer i…

xpositivityx:
…n bundle to omit from publish if it does not exist

### DIFF
--- a/pkg/application/parse.go
+++ b/pkg/application/parse.go
@@ -22,11 +22,27 @@ func Parse(path string, overrides map[string]interface{}) (*bundle.Bundle, error
 		log.Fatalf("Unmarshal: %v", err)
 	}
 
+	applyAppBlockDefaults(app)
+
 	if overrideErr := applyOverrides(app, overrides); overrideErr != nil {
 		return nil, overrideErr
 	}
 
 	return app, nil
+}
+
+func applyAppBlockDefaults(b *bundle.Bundle) {
+	if b.App != nil {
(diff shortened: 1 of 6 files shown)


pkg/bundle/types.go
```diff
@@ -21,12 +21,20 @@ type Bundle struct {
 	Params      map[string]interface{} `json:"params" yaml:"params"`
 	Connections map[string]interface{} `json:"connections" yaml:"connections"`
 	UI          map[string]interface{} `json:"ui" yaml:"ui"`
-	App         AppBlock               `json:"app" yaml:"app"`
+	App         *AppBlock              `json:"app" yaml:"app"`
```
xpositivityx, line 24: Made this a pointer so that in the marshalling on publish it will be completely omitted if there was no app block in the massdriver.yaml

pkg/application/parse.go
```diff
@@ -22,13 +22,29 @@ func Parse(path string, overrides map[string]interface{}) (*bundle.Bundle, error
...
+	applyAppBlockDefaults(app)
+
 	if overrideErr := applyOverrides(app, overrides); overrideErr != nil {
 		return nil, overrideErr
 	}
 
 	return app, nil
 }
 
+func applyAppBlockDefaults(b *bundle.Bundle) {
```
xpositivityx, line 34: If there was any part of an app spec, make sure all values have a default if they are not set.

coryodaniel (approved):

PR #153, opened 2023-01-11
remove unrelated code

WillBeebe:
This must be what you meant. Sorry man, I always delete this myself! I remember looking at this when I checked in this template and thinking huh, when would I ever use that. I didn't write it and always delete it but my name's on it so here ya go.


pkg/bundle/templates/terraform/massdriver.yaml
```diff
@@ -63,11 +63,3 @@ artifacts:
 ui:
   ui:order:
     - "*"
-  ## set top level ui:wiget
```
coryodaniel, line 66: Add above the `ui` block: `# learn more about widgets & fields: https://docs.massdriver.cloud/bundles/custom-widgets-and-fields

WillBeebe:
Oh Daniel @coryodaniel feedback taken.

coryodaniel (approved):

PR #161, opened 2023-02-09
add docker platform flag

WillBeebe:


xpositivityx (commented):
What is the use case for skipping push? Pushing is the hard part. Users should probably just use docker build if they don't want to push. We add no value over the existing tooling and we require a bunch of extra args.

WillBeebe:
> What is the use case for skipping push? Pushing is the hard part. Users should probably just use docker build if they don't want to push. We add no value over the existing tooling and we require a bunch of extra args.

My idea was to test the build without pushing the image remotely. I'll yank the skip logic.

xpositivityx (approved):

PR #167, opened 2023-02-27
Mass app new require description

chrisghill:
This is fixing some issues I've scene customers run into when setting up apps.

Apps are likely stored in the same directory as the application code itself, so a `massdriver` subdirectory (`app/massdriver/massdriver.yaml`) makes more sense than using the application name as the subdirectory (`app/app/massdriver.yaml`).

If the description field is empty it gets rejected by the API, so we should prevent empty descriptions when creating a new app/bundle.

### DIFF
--- a/pkg/application/prompt.go
+++ b/pkg/application/prompt.go
@@ -81,8 +81,16 @@ func getAccessLevel(t *template.Data) error {
 }
 
 func getDescription(t *template.Data) error {
+	validate := func(input string) error {
+		if len(input) == 0 {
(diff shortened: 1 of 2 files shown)


WillBeebe:
Thank you! I've been kicking these headaches down the road.

WillBeebe (approved):
Looks pretty shippable to me, good work.

WillBeebe:
> Looks pretty shippable to me, good work.

@coryodaniel annoying ping to check that this is in V2 or slated.

Issue #152, opened 2023-01-09
Replace file prompt in `mass app new` not working


mclacore:
![cli-bug](https://user-images.githubusercontent.com/8037512/211378401-9fcf5c15-33c9-4657-ac91-812c85273fdb.png)

When I run `mass app new` and go through the prompts, if `README.md` already exists, it'll ask me if I want to overwrite the existing file. If I enter `n` or `N`, it still overwrites the file.

PR #168, opened 2023-03-08
liquid support

coryodaniel:
Still need to skip subdirs or add `{% raw %}` blocks to templates.

Also closes: https://github.com/massdriver-cloud/massdriver-cli/issues/152


akasprzok (approved):
Thanks for taking this on, would've taken me forever w/o any CLI context.

akasprzok:
> Still need to skip subdirs or add `{% raw %}` blocks to templates.

On the massdriver side I currently only render "massdriver.yaml" and "README.md".

Not sure what the best option here is overall.

Raw tags seems "cleaner" from an implementation perspective but does complicate the templates in general.

pkg/bundle/generate.go
```diff
@@ -30,6 +34,7 @@ func Generate(data *template.Data) error {
 	data.RepoName = fmt.Sprintf("massdriver-cloud/%s", data.Name)
 	data.RepoNameEncoded = strings.ReplaceAll(data.RepoName, "/", "%2F")
 
-	templateDir, _ := fs.Sub(fs.FS(templatesFs), path.Join("templates", data.TemplateName))
-	return template.RenderEmbededDirectory(templateDir, data)
+	data.TemplateSource = cache.AppTemplateCacheDir()
+	templatePath := path.Join(data.TemplateSource, data.TemplateName)
+	return template.RenderDirectory(templatePath, data)
```
coryodaniel, line 39: this is using app rendering now. Added tf as a template there.

pkg/bundle/generate_test.go
```diff
@@ -1,62 +1,3 @@
 package bundle_test
 
-import (
```
coryodaniel, line 3: noop, test dupe of template.Render now.

coryodaniel:
Ended up implementing as templating anything in the root dir of the template, and copying any subdirs. This allows us to use rendering in operator files, etc. note [here](https://github.com/massdriver-cloud/massdriver-cli/pull/168/files#diff-677d343ec8027c2775ceaac5f0ca4bf1e2af203d41c8cdb2d77d8bb32d8e5e78R11)

pkg/template/files.go
```diff
@@ -1,81 +1,24 @@
...
-}
-
-func promptAndWrite(file []byte, data *Data, outputPath string) error {
-	tmpl, errTmpl := template.New("tmpl").Delims(OpenPattern, ClosePattern).Parse(string(file))
-	if errTmpl != nil {
-		return errTmpl
-	}
+	var bindings map[string]interface{}
+	inrec, _ := json.Marshal(data)
+	json.Unmarshal(inrec, &bindings)
```
coryodaniel, line 15: needed a map string interface, had a struct.

xpositivityx (approved):

PR #172, opened 2023-03-13
Strongly typed connections

akasprzok:
instead of using a map, we strongly type connections as

```
Connection {
    Name
    ArtifactDefinition
}
```

### DIFF
--- a/pkg/template/types.go
+++ b/pkg/template/types.go
@@ -1,16 +1,21 @@
 package template
 
 type Data struct {
-	Name            string                 `json:"name"`
-	Description     string                 `json:"description"`
-	Access          string                 `json:"access"`
-	Location        string                 `json:"location"`
-	TemplateName    string                 `json:"templateName"`
-	TemplateSource  string                 `json:"templateSource"`
-	OutputDir       string                 `json:"outputDir"`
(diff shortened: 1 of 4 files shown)


pkg/template/types.go
```diff
@@ -1,16 +1,21 @@
...
-	RepoNameEncoded string                 `json:"repoNameEncoded"`
+	Name            string       `json:"name"`
+	Description     string       `json:"description"`
+	Access          string       `json:"access"`
+	Location        string       `json:"location"`
+	TemplateName    string       `json:"templateName"`
+	TemplateSource  string       `json:"templateSource"`
+	OutputDir       string       `json:"outputDir"`
+	Type            string       `json:"type"`
+	Connections     []Connection `json:"connections"`
```
xpositivityx, line 12: this field is also used to create the connections schema. We should check the tests for build and make sure these changes are reflected in the test data.

akasprzok: Which connections schema? I don't see any build tests either. Can you point me in the right direction?

akasprzok: This change only impacts YAML generation, not any of the other flows.

xpositivityx (approved):

